PR #15, opened 2015-05-01
Switch to QT5

crrodriguez:
Only minimal changes to the build process required.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,8 +9,10 @@ set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake/")
 
 include(GNUInstallDirs)
 
-find_package(Qt4 REQUIRED QtCore QtGui QtXmlPatterns)
-include(${QT_USE_FILE})
+find_package(Qt5Widgets REQUIRED)
+find_package(Qt5XmlPatterns REQUIRED)
+
+add_definitions(-DQT_DISABLE_DEPRECATED_BEFORE=0)
 
 # This conditional is here to avoid checking openscap version if user supplies
 # custom openscap parameters.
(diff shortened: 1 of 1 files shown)


mpreisler:
Hi,

thanks a lot for your contribution! Did you notice any differences at runtime?

The reason why SCAP Workbench remains Qt4 is because RHEL6 and RHEL7 are stuck with Qt4. Is SUSE moving to Qt5?

Maybe it makes sense to move SCAP Workbench master to Qt5 and leave maintenance branches in Qt4. I will have to think about this a bit more.

crrodriguez:
> Is SUSE moving to Qt5?
Well, not only SUSE but everybody has to.. QT4 EOL is due by the end of 2015.

I did not find any problem with the resulting executable.

mpreisler:
The enterprise distros still support Qt3, I am sure Qt4 will survive for a long long time.

But you are right, we should move to Qt5 in master.

mpreisler:
Qt.io suggests to migrate to Qt5 - http://blog.qt.io/blog/2015/05/26/qt-4-8-7-released/
> Many users have already moved their active projects to Qt 5 and we encourage also others to do so. With a high degree of source compatibility, we have ensured that switching to Qt 5 is smooth and straightforward. It should be noted that Qt 4.8.7 provides only the basic functionality to run Qt based applications on Mac OS X 10.10, full support is in Qt 5.

> Qt 4.8.7 is planned to be the last patch release of the Qt 4 series. Standard support is available until December 2015, after which extended support will be available. We recommend all active projects to migrate to Qt 5, as new operating systems and compilers with Qt 4.8 will not be supported. If you have challenges migrating to Qt 5, please contact us or some of our service partners for assistance.

From this I conclude that we should move to Qt5 in master. I will check availability of Qt5 in various distros and report back.

mpreisler:
I am reluctant to merge this because Qt5 is not readily available in RHEL6 and RHEL7.

On the other hand we can just keep v1-1 for distros with Qt4 and develop with Qt5 on master.

mpreisler:
Decided to use Qt5 in master, v1-1 will be the last branch that supports Qt4.

Thanks for your contribution!

PR #22, opened 2015-06-28
fix two minor bugs in scap-workbench.8 manpage

finkr:
The manpage has some minor bugs:

* Fix manpage title (name), so the whatis command work properly.
* Fix hyphen where a minus sign was intended


mpreisler:
Thanks for your contribution!

Issue #28, opened 2015-07-10
SCAP Workbench 1.1.0 - Inconsisten benchmark names in the "Open SCAP Security Guide" dialog


iankko:
Description of problem:

SCAP Workbench of version 1.1.0 and earlier does not use a consistent way in order to name
button labels in the "Open SCAP Security Guide" dialog (some benchmarks are titled with first letter capitalized, some benchmarks labels are completely in lowercase).

Version:
scap-workbench-1.1.0-2.el7.centos.x86_64 from isimluk's COPR repository

How reproducible:
Always

Steps to Reproduce:
* run scap-workbench with latest SCAP Security Guide 0.1.24 version installed,
* notice the benchmark labels in the displayed "Open SCAP Security Guide" dialog

Current result:
See the attached image -- some labels have first letter capitalized, while the others are completely in lowercase
![scap-workbench-1 1 0-screenshot](https://cloud.githubusercontent.com/assets/8414918/8618257/98f66852-270c-11e5-8b10-0927e58601a5.png)

Expected result:
Uniform approach when displaying benchmark names - either all labels to be in lowercase, or all of them to have first letter capitalized (maybe 'JRE' is an exception here, and it should have all the letters capitalized).

Additional information:
Also notice the way how SCAP Security Guide logo is displayed -- instead of whole logo, there's top and bottom part of it greyed out. I can file a separate ticket for this if required.

mpreisler:
There are 2 issues that should be fixed as part of this ticket.
* Support more content folders and rewrite them into nicer names - e.g.: `CentOS6` instead of `centos6`, `JRE` instead of `jre`, ... - this is relatively trivial
* Pack all the buttons into a scrollarea so that it scales better. When this dialog was created I didn't envision `SSG` shipping so much content, I greatly underestimated the speed at which the project progressed in the past few months :-) - this requires a bit of `Qt4` knowledge, I am willing to mentor

iankko:
Thanks, Martin.

Could within the proposed changes be assured that the widget displaying SCAP Security Guide logo is large enough to hold the whole logo? IOW ensure it's not cut off like in actual version.

Thanks, Jan.

mpreisler:
> Could within the proposed changes be assured that the widget displaying SCAP Security Guide logo is large enough to hold the whole logo? IOW ensure it's not cut off like in actual version.

[...]

PR #31, opened 2015-08-14
https://github.com/OpenSCAP/scap-workbench/issues/28

ybznek:
screenshot_2015-08-14_18-53-28

The pullrequest:
 - Add scrollbar
 - Make benchmark names consistent
 - Divide benchmark into two "groups" preffered/non-preffered

### DIFF
--- a/src/SSGIntegrationDialog.cpp
+++ b/src/SSGIntegrationDialog.cpp
@@ -90,11 +90,13 @@ void SSGIntegrationDialog::scrapeSSGVariants()
 
         QString label = name;
 
+        bool favorite = false;
         // Make the label nicer for known variants
-        if (label.startsWith("rhel")) // use RHEL instead of rhel
+        if (label.startsWith("rhel")) { // use RHEL instead of rhel
             label = name.toUpper();
+            favorite = true;
 
-        else if (label.startsWith("centos")) // use CentOS instead of centos
+        } else if (label.startsWith("centos")) // use CentOS instead of centos
             label.replace(0, 6, "CentOS");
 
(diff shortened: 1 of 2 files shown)


mpreisler:
Hi,
thanks for your contribution! Please try to match the code style of the project. Curly braces go on a new line.

Other than that I see no major issues.

Could you post a screenshot of how this looks like?

mpreisler:
#28

src/SSGIntegrationDialog.cpp
```diff
@@ -90,15 +90,33 @@ void SSGIntegrationDialog::scrapeSSGVariants()
...
+
+        else
+            label[0] = label[0].toUpper(); // Capitalize first letter
+
+        QPushButton* button = new QPushButton(label, mUI.content);
         button->setProperty("ssg_variant", QVariant(name));
-        mUI.variants->layout()->addWidget(button);
+
+        if (favorite)
+            mUI.variantsLayout->insertWidget(1, button); // insert button before text and divider
```
mpreisler: Does this mean that favorites end up being in reverse-sorted order and everything else is in sorted order?

ybznek: Yes... I understand favorites as some SET of preferred options. I don't know WHICH items should be preferred and HOW we want to order them.
Current implementation is the easiest possible.

mpreisler: In my opinion it looks odd that the order is reversed. How about remembering the last index of the favorites set and always appending?

mpreisler:
Could you please add `Fedora` to favorites?

mpreisler:
Thx

PR #32, opened 2015-08-14
Fix root commands highlight in README.md

ybznek:
Preview: https://github.com/ybznek/scap-workbench/blob/v1-1-readme/README.md


mpreisler:
Thx

Issue #25, opened 2015-07-07
Non intuitive "Close currently opened file" dialog text


dahaic:
Hello,
I have two proposals for changes to that dialog:

1. Usually when such dialog is presented to the user, it contains positive type of question, like:
`Do you want to proceed?` with explanation what proceeding consists of provided beforehand. Asking if user wants to close file feels like negative question and thus a bit harder to understand.

2. User may be more interested in identifying which file is being closed, than what file (s)he wants to open, as (s)he selected it just few seconds earlier.

b-long:
:+1: I agree, certainly a fan of anything to improve user experience

mpreisler:
Fixed by #38

PR #38, opened 2015-08-19
Non intuitive "Close currently opened file" dialog text

ybznek:
https://github.com/OpenSCAP/scap-workbench/issues/25

### DIFF
--- a/include/MainWindow.h
+++ b/include/MainWindow.h
@@ -28,6 +28,7 @@
 #include <QMainWindow>
 #include <QThread>
 #include <QMenu>
+#include <QMessageBox>
 
 extern "C"
 {
@@ -262,6 +263,13 @@ class MainWindow : public QMainWindow
          * @see refreshSelectedRulesTree
          */
         unsigned int getSelectedRulesCount();
+
+        /**
+         * @brief Ask user to proceed with closing old file due to openning of new one
+         *
+         * @return user answer
+         */
+        QMessageBox::StandardButton openNewFileQuestionDialog(const QString& newFilepath);
 
     private slots:
         /**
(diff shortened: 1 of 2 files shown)


ybznek:
First commit merge two duplicate codes. Second commit is coming.

src/MainWindow.cpp
```diff
@@ -1406,3 +1397,11 @@ void MainWindow::aboutQt()
 {
     QMessageBox::aboutQt(this);
 }
+
+
+QMessageBox::StandardButton MainWindow::openNewFileQuestionDialog(QString oldFilepath) {
```
mpreisler: code style...

src/MainWindow.cpp
```diff
@@ -1406,3 +1397,11 @@ void MainWindow::aboutQt()
...
+    return QMessageBox::question(this, QObject::tr("Close currently opened file?"),
+                               QObject::tr("Opened file '%1' has to be closed before opening another file.\n\n"
```
mpreisler: strange indentation

src/MainWindow.cpp
ybznek: specifically what?

mpreisler: code style - newline before scope opening brace
pass by value - the `QString oldFilepath`

ybznek: Ah, ok.
QString has COW.. But passing as const & could be faster.

include/MainWindow.h
```diff
@@ -263,6 +264,13 @@ class MainWindow : public QMainWindow
          */
         unsigned int getSelectedRulesCount();
 
+        /**
+         * @brief Ask user to proceed with closing old file due to openning of new one
+         *
+         * @return user answer
+         */
+        QMessageBox::StandardButton openNewFileQuestionDialog(const QString &oldFilepath);
```
mpreisler: The SCAP Workbench codebase uses the `const T& var` convention instead of `const T &var`.

PR #39, opened 2015-08-20
Add padding to cell of Diagnostics output

ybznek:
Add padding to diagnostics output cell
Before
--------
after

screenshot_2015-08-20_18-13-08

### DIFF
--- a/src/DiagnosticsDialog.cpp
+++ b/src/DiagnosticsDialog.cpp
@@ -131,7 +131,7 @@ void DiagnosticsDialog::pushMessage(MessageSeverity severity, const QString& ful
 
     std::cerr << stime << " | " << strSeverity.toUtf8().constData() << " | " << fullMessage.toUtf8().constData() << std::endl;
     mUI.messages->append(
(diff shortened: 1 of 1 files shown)


src/DiagnosticsDialog.cpp
```diff
@@ -131,7 +131,7 @@ void DiagnosticsDialog::pushMessage(MessageSeverity severity, const QString& ful
 
     std::cerr << stime << " | " << strSeverity.toUtf8().constData() << " | " << fullMessage.toUtf8().constData() << std::endl;
     mUI.messages->append(
-        QString("<table><tr><td><pre>%1 </pre></td><td style=\"background: %2\"><pre>%3 </pre></td><td>%4</td></tr></table>\n")
+        QString("<table><tr><td style=\"padding:5px 0px 0px 5px\"><pre>%1 </pre></td><td style=\"background: %2; padding:5px\"><pre>%3 </pre></td><td>%4</td></tr></table>\n")
```
mpreisler: With the padding added, is the trailing space in `<pre>` necessary?

ybznek: I think, You have the same width of the cell due to trailing spaces. Without them you would not get the same size in every line. Maybe it would be better specify width of cell only in css.

Issue #27, opened 2015-07-07
empty error message when ending tag missing


dahaic:
When trying to open xccdf file referencing malformed oval file, where closing tag `</definitions>` is missing, error message workbench presents to the user is empty.

Output of `oscap oval validate test-rhel7-oval.xml` (exit code 1):
```
test-rhel7-oval.xml:57: parser error : Opening and ending tag mismatch: definitions line 15 and oval_definitions
</oval_definitions>
                   ^
test-rhel7-oval.xml:58: parser error : Premature end of data in tag oval_definitions line 1

^
```

Output of `scap-workbench`:
```
Error while opening file. There was a problem with ScanningSession! Failed to reload session. OpenSCAP error message: 
```

mpreisler:
Solved by @ybznek in #47

PR #41, opened 2015-08-26
Show LibXML errors in Diagnostics Dialog

ybznek:
Add LibXML errors to diagnostics dialog according to
https://github.com/OpenSCAP/scap-workbench/issues/27

### DIFF
--- a/src/LibXmlErrorGuard.cpp
+++ b/src/LibXmlErrorGuard.cpp
@@ -39,7 +39,7 @@ const QString& LibXmlErrorGuard::getMessage() const
    return errorMessage;
 }
 
-bool LibXmlErrorGuard::isEmpty()
+bool LibXmlErrorGuard::isEmpty() const
 {
     return errorMessage.isEmpty();
 }
--- a/src/LibXmlErrorGuard.cpp
+++ b/src/LibXmlErrorGuard.cpp
@@ -1,0 +1,61 @@
+/*
+ * Copyright 2013 Red Hat Inc., Durham, North Carolina.
+ * All Rights Reserved.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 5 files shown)


ybznek:
Still not ready to Merge. It need some more "beautiful" output (and don't destroy xml tags in html output)

screenshot_2015-08-26_16-11-12

src/LibXmlErrorGuard.cpp
```diff
@@ -0,0 +1,42 @@
...
+bool LibXmlErrorGuard::isEmpty()
+{
+    return errorMessage.isEmpty();
+}
+
+void LibXmlErrorGuard::libxmlErrorCallback(QString* message, const char* format, ... )
+{
+    try
+    {
+      va_list args;
```
ybznek: Uff, bad indentation again...

src/LibXmlErrorGuard.cpp
```diff
@@ -0,0 +1,42 @@
...
+    // set default libxml error handler
+    xmlSetGenericErrorFunc(NULL, (xmlGenericErrorFunc)NULL);
+}
+
+const QString& LibXmlErrorGuard::getMessage() const
+{
+   return errorMessage;
+}
+
+bool LibXmlErrorGuard::isEmpty()
```
mpreisler: const

src/LibXmlErrorGuard.cpp
```diff
@@ -0,0 +1,42 @@
...
+    {
+        va_list args;
+        va_start(args, format);
+        QString newMessage = QString().vsprintf(format, args);
+        va_end(args);
+        message->append(newMessage);
+    }
+    catch(...)
+    {
+        ; // Possible exceptions aren't safe in C callback
```
mpreisler: I don't see how exceptions can happen in the block, maybe I am missing something. Qt doesn't throw.

ybznek: I think "new" somewhere inside of these functions can throw exception, or cannot?

mpreisler: It can throw `std::bad_alloc` AFAIK but if that happens all is lost anyway :-)

ybznek:
screenshot_2015-08-26_18-20-43

Any suggestions?

mpreisler:
Looks great.

mpreisler:
Discussed this with @ybznek , the fix has to go into OpenSCAP and not SCAP Workbench.

ybznek:
Probably... It will need some fix in Workbench too (printing preformated xml error message/escape xml before output/..) But I can cherry-pick these changes...

PR #43, opened 2015-09-01
Fix strange characters in report after remote scan

ybznek:
Steps to reproduce:
- select remote scan (Linux->Linux) (you can select your localhost)
- scan
- show report -> You will see 'Â' in front of addresses (and in some other parts of document)

screenshot_2015-09-01_13-36-58

@msrubar found this bug

### DIFF
--- a/src/ProcessHelpers.cpp
+++ b/src/ProcessHelpers.cpp
@@ -163,8 +163,8 @@ void SyncProcess::run()
 
     mRunning = false;
 
-    mStdOutContents = QString::fromUtf8(process.readAllStandardOutput());
-    mStdErrContents = QString::fromUtf8(process.readAllStandardError());
+    mStdOutContents = QString::fromLocal8Bit(process.readAllStandardOutput());
+    mStdErrContents = QString::fromLocal8Bit(process.readAllStandardError());
 
     // TODO: We are duplicating data here!
     mDiagnosticInfo += "stdout:\n===============================\n" + QString(mStdOutContents) + QString("\n");
--- a/src/ProcessHelpers.cpp
+++ b/src/ProcessHelpers.cpp
@@ -163,8 +163,8 @@ void SyncProcess::run()
 
     mRunning = false;
 
-    mStdOutContents = process.readAllStandardOutput();
-    mStdErrContents = process.readAllStandardError();
+    mStdOutContents = QString::fromUtf8(process.readAllStandardOutput());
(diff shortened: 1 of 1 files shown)


ybznek:
It probably need some testing with other platforms, maybe it can break something.

@mpreisler What do you think? Do you have some testing suite or sth like that?

mpreisler:
Thanks for spotting and fixing this!

Unfortunately we have no test suite. I will check this out later today. Just watching the source changes I don't think it should break anything.

src/ProcessHelpers.cpp
```diff
@@ -163,7 +163,7 @@ void SyncProcess::run()
 
     mRunning = false;
 
-    mStdOutContents = process.readAllStandardOutput();
+    mStdOutContents = QString::fromUtf8(process.readAllStandardOutput());
     mStdErrContents = process.readAllStandardError();
```
mpreisler: Shouldn't the same fix be applied for mStdErrContents?

ybznek: maybe, but I don't expect non ascii characters on stderr.

mpreisler:
We might want to add to `QT_NO_CAST_FROM_ASCII` define to `CMakeLists.txt`, see http://doc.qt.io/qt-4.8/qstring.html#QString-9

This would prevent future mistakes similar to this.

mpreisler:
One more thing, is `QProcess::readAllStandardOutput` guaranteed to always return UTF8? Even on systems with `cp1250` locale or other 8bit encoding locale? Wouldn't http://doc.qt.io/qt-4.8/qstring.html#fromLocal8Bit be a better match for this use-case?

ybznek:
Thats is reason why I've mentioned that it can break something.
I don't know set of supported platforms and their encoding, but fromLocal8Bit could solve it.

mpreisler:
`fromLocal8Bit` solves it unless `readAllStandardOutput` can return `utf16` or `utf32`. I am not sure about that.

ybznek:
QT_NO_CAST_FROM_ASCII cause multiple issues like this

OscapCapabilities.cpp:35:14: error: within this context mVersion = "Unknown";

error: within this context
     if (versionGreaterOrEqual(mVersion, "0.9.5"))

Do we really want to define the contant?

mpreisler:
You are right, let's not do the defines in this pull request. I think they could be done in master if we get any spare time to lower tech debt.

PR #47, opened 2015-09-03
Show LibXML errors in Diagnostics Dialog

ybznek:
rnv6


ybznek:
Maybe there could be still some warnings while xml is loading which don't cause error and are not printed to user.

ybznek:
It need https://github.com/OpenSCAP/openscap/pull/151 and optionally https://github.com/OpenSCAP/openscap/pull/150.

mpreisler:
Just to be clear, it needs those 2 pull requests to get nice errors but if I merge this workbench 1.1 will continue to work on all openscap versions 1.2.0 and higher. The errors won't be as nice but it will compile and work.

Issue #24, opened 2015-06-29
Search works only in selected sub-trees


zemb:
Let's say I use RH CCP profile from RHEL-7 XCCDF.
When I search for "encrypt", it finds "Encrypt Transmitted Data Whenever Possible" and "Encrypt Partitions".
But if I deselect "Introduction" the "Encrypt Transmitted Data Whenever Possible" under "Introduction" is not found when doing the search.

ybznek:
@mpreisler I think it is caused by `disabled` tree elements - cannot be focused . Did you have any reason why they are disabled and not only `unchecked`?

ybznek:
Fixed by https://github.com/OpenSCAP/scap-workbench/issues/73

PR #73, opened 2016-07-31
Fix search of disabled items

ybznek:
It is the most clean & working solution I have found.

- Settings of `setForeground()` was very slow from some reason (https://bugreports.qt.io/browse/QTBUG-10015) and `setUpdatesEnabled();` didn't help

- Set item temporary to "enabled" worked only sometimes

https://github.com/OpenSCAP/scap-workbench/issues/24


mpreisler:
ACK

Issue #67, opened 2016-06-29
Tailoring of non-interactive value


ybznek:
Value[@interactive] is 0 in default

Workbench currently allows to edit all values - non-interactive, too - e.g. multiline scripts (in single line text edit)

https://scap.nist.gov/specifications/xccdf/xccdf_element_dictionary.html#valueType

ybznek:
Should be fixed by https://github.com/OpenSCAP/scap-workbench/issues/77

PR #77, opened 2016-08-24
Don't allow edit values with changes-prohibited

ybznek:
https://github.com/OpenSCAP/scap-workbench/issues/67
According to SSG, it use `prohibitChanges` instead of `interactive`
```
<Value id="xccdf_org.ssgproject.content_value_function_fix_audit_syscall_rule" hidden="true" prohibitChanges="true" operator="equals" type="string">
```
Do you have better idea how to visually show "changes prohibited"? (otherwise it should be ready to merge)
screenshot_2016-08-24_14-45-19

### DIFF
--- a/src/TailoringDockWidgets.cpp
+++ b/src/TailoringDockWidgets.cpp
@@ -212,17 +212,28 @@ void XCCDFItemPropertiesDockWidget::refresh()
                     break;
             }
 
-            struct xccdf_value_instance_iterator* it = xccdf_value_get_instances(value);
-            while (xccdf_value_instance_iterator_has_more(it))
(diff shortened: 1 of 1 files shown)


iankko:
@ybznek 
Yes, we use ```prohibitChanges``` since we don't want the code of the remediation functions to be possible to be modified by the user. The [```prohibitChanges```](https://scap.nist.gov/specifications/xccdf/xccdf_element_dictionary.html#itemType) attribute seemed the most appropriate candidate to me how to reach ```If benchmark producers should prohibit changes to this item during tailoring. An author should use this when they do not want to allow end users to change the item. ```

The proposed change looks good to me (from UX PoV).

PR #79, opened 2016-09-13
RHBZ#1368527 Add support for --version and --help

yuumasato:
Arguments -v and --version prints the version of SCAP Workbench,
and arguments -h and --help prints the usage of SCAP Workbench.

This is the same funcionality added  by
commit 9d821197d5c13e058dbdc5b8a6a03839e5817e6a, but not
using QCommandLineParser, which is only available in Qt5.

### DIFF
--- a/include/Application.h
+++ b/include/Application.h
@@ -60,6 +60,21 @@ class Application : public QApplication
          */
         void browseForContent();
 
+        /**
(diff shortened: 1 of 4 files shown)


jan-cerny:
add to whitelist

jan-cerny:
ok to test

PR #81, opened 2016-09-21
Fix tailoring file path of dry run

yuumasato:
RHBZ#1368516

Provide actual file path of tailoring file when using dry run feature.
If tailoring is not saved, a temporary tailoring path is provided.

Also, provide paths with white spaces protected by double quotes.

### DIFF
--- a/src/OscapScannerRemoteSsh.cpp
+++ b/src/OscapScannerRemoteSsh.cpp
@@ -121,9 +121,8 @@ QStringList OscapScannerRemoteSsh::getCommandLineArgs() const
     }
     else
     {
-        QString userTailoringFile = mSession->getUserTailoringFilePath();
         args += buildEvaluationArgs(mSession->getOpenedFilePath(),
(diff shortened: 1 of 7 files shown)


src/OscapScannerRemoteSsh.cpp
```diff
@@ -121,8 +121,9 @@ QStringList OscapScannerRemoteSsh::getCommandLineArgs() const
     }
     else
     {
+        QString userTailoringFile = mSession->getUserTailoringFilePath();
         args += buildEvaluationArgs(mSession->getOpenedFilePath(),
-            mSession->hasTailoring() ? mSession->getTailoringFilePath() : QString(),
+            mSession->hasTailoring() ?  ( userTailoringFile.isEmpty() ? mSession->getTailoringFilePath() : userTailoringFile ): QString(),
```
ybznek: IMHO, you could create ScanningSession method to wrap these double ternary operators

yuumasato: Thank you for the suggestion. I followed the idea of a wrap method and ended up changing the use of getUserTailoringFilePath().

mpreisler:
Awesome, ACK!

PR #83, opened 2016-10-03
Improve SSGIntegrationDialog UX

yuumasato:
These changes attempt to improve UX of SSGIntegrationDialog.

- the button list has been replaced by a combo box,
- Instructions in the dialog have changed,
- and workflow of window closing have changed, for example closing SSGIntegrationDialog at startup of application closes workbench.


mpreisler:
While this does introduce an extra click for experienced SCAP Workbench users I think it improves the experience for new users because the UI elements are more conventional.

PR #85, opened 2016-10-06
Improve tailoring UX part 1

yuumasato:
Some improvements in the screens and message boxes used during the tailoring of a profile.


mpreisler:
ACK

PR #86, opened 2016-10-14
Improve some more UX in main window

yuumasato:
- Added mnemonics to save results format options
- Added mnemonics to options in menu
- Added save shortcut description into save results button tooltip
- Added labels to remote machine related fields
- Switched recent hosts button menu for a combo box

### DIFF
--- a/ui/ResultViewer.ui
+++ b/ui/ResultViewer.ui
@@ -34,7 +34,7 @@
          </font>
         </property>
         <property name="toolTip">
-         <string>Saves results of this evaluation.</string>
+         <string>Saves results of this evaluation. (Ctrl+S)</string>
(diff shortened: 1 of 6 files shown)


ui/ResultViewer.ui
```diff
@@ -34,7 +34,7 @@
          </font>
         </property>
         <property name="toolTip">
-         <string>Saves results of this evaluation.</string>
+         <string>Saves results of this evaluation. (Ctrl+S)</string>
```
ybznek: I am not UX guy, but I think we could have Save results option in main menu and it could be grayed out with no results are available.

mpreisler:
ACK

PR #88, opened 2016-10-14
Improve UX of about window

yuumasato:
Links in the description were made clickable.


yuumasato:
I think the easier way to remove the icon from OK button is by removing all icons from the application, done in PR #89.
An other approach would be by designing the about dialog manually.

ybznek:
changes LGTM, but personally I don't think we still use https://fedorahosted.org/scap-workbench/

cc @mpreisler

mpreisler:
Good catch @ybznek 

Please change the URL to: `https://www.open-scap.org/tools/scap-workbench/`

yuumasato:
URL updated.

mpreisler:
ACK

PR #89, opened 2016-10-14
Remove icons from every dialog box

yuumasato:
This change removes icons from every button in dialog boxes, including
the default Open and Save file dialogs.

### DIFF
--- a/ui/MainWindow.ui
+++ b/ui/MainWindow.ui
@@ -2,6 +2,9 @@
 <ui version="4.0">
  <class>MainWindow</class>
  <widget class="QMainWindow" name="MainWindow">
+  <property name="styleSheet">
+   <string notr="true">dialogbuttonbox-buttons-have-icons: 0;</string>
+  </property>
   <property name="enabled">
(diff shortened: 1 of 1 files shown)


yuumasato:
This may be exaggerated, but this way all buttons in dialogs across the workbench are consistent now.
Including default Open/Save File dialogs.

See the before/after screenshot:
save_customization_icons

ybznek:
Actually I can still see icons in dialogs on my XFCE.
image
I don't think that it is problem in general, but i expect that you don't want to have such icons.

mpreisler:
We are using the system file dialogs to better blend into the system desktop interface. Did the UX review recommend to alter these dialogs? I thought using system dialogs is the de-facto UX friendly way of doing things :-)

yuumasato:
The UX review did not have a screenshot of these system file dialogs, they didn't see them...

But their saying about icons is:
> Standard presentation of action buttons in modal dialogs is without icons

mpreisler:
Let's use system dialogs as they are.

PR #90, opened 2016-10-17
Do not discard colon when reading dowload url

yuumasato:
When fetching remote content, openscap will inform the scap-workbench
about resources being downloaded. The colon in the http URL of the
resource was being removed


mpreisler:
ACK

PR #92, opened 2016-12-05
[DO NOT MERGE] Show number of selected rules in profile listing

### LINKED ISSUE #40: [RFE] Hide 'default' profile if it does not contain some selected rules
`scap-workbench` currently provides listing of profiles present in the benchmark, that can be selected to be used to scan the system in question. The `default` profile is exceptional in this context. Exceptional in the sense, it's not present in every benchmark, and it's required by the SCAP standard to be provided "artificially" by the scanner. The motivation behind this requirement being that `default` profile it's intended to contain all rules (if no more specific XCCDF profile is selected).

Though in the context of SCAP Security Guide project the `default` profile is not used (the `default` profiles contains no rules selected). Therefore for SCAP Security Guide purpose the `default` profile is of very limited use (one use case coming to mind is to select `default` profile to have empty benchmark, then use tailoring to select just couple of rules to be scanned).

Therefore this RFE is to request the `default` profile to be shown in the listing of available `scap-workbench` profiles only in the case the `default` profile provided from the content truly contains some of the rules being selected (e.g. it would not be displayed if it does not contain some selected rules).

Thank you, Jan.

yuumasato:
To avoid confusion on the semantics of the `default` profile, we show
that this default profile has no selected rules.

Fixes #40


src/MainWindow.cpp
```diff
@@ -649,7 +656,8 @@ void MainWindow::refreshProfiles()
 #if (QT_VERSION >= QT_VERSION_CHECK(4, 4, 0))
         mUI.profileComboBox->insertSeparator(mUI.profileComboBox->count());
 #endif
-        mUI.profileComboBox->addItem(QObject::tr("(default)"), QVariant(QString::Null()));
+        // In workbench, there is always a default profile with no rules selected
+        mUI.profileComboBox->addItem(QObject::tr("(default) (0)"), QVariant(QString::Null()));
```
mpreisler: This is not correct, `(default)` profile can have selected rules if any of the Rules have the `@selected` attribute set to `true`. You can use the same code with `xccdf_policy_get_selected_rules_count` to check the default profile.

yuumasato:
2e917de relies on a fix suggested to OpenSCAP: https://github.com/OpenSCAP/openscap/pull/607

mpreisler:
We should change https://github.com/OpenSCAP/scap-workbench/blob/v1-1/CMakeLists.txt#L19 as part of this PR to prevent potential segfaults.

yuumasato:
I'm trying to check version of openscap at compile time, rather than at runtime, and still trying to keep things simple.

I couldn't find a way to obtain the version of a library specified through OPENSCAP_LIBRARIES.
So it fails when an openscap library is specified with OPENSCAP_LIBRARIES and OPENSCAP_INCLUDE_DIRS. 

Any suggestion?

mpreisler:
I just noticed that you are suggesting this for v1-0, why?

mpreisler:
> I couldn't find a way to obtain the version of a library specified through OPENSCAP_LIBRARIES.
> So it fails when an openscap library is specified with OPENSCAP_LIBRARIES and OPENSCAP_INCLUDE_DIRS.
> 
> Any suggestion?

I looked into this and you are right, there are no easy ways to get it. We have a couple of options:
- do it as a runtime check, it's dirty but if commented I think this is acceptable
- if we can get a compile time version use that, if we can't, assume we can't get selected rules for default profile

yuumasato:
> I just noticed that you are suggesting this for v1-0, why?
It's the oldest branch that can receive the enhancement. But bug fixes goes to the oldest branch, right?

yuumasato:
Closing this PR an opening one for branch v1-1, see #98.

PR #93, opened 2016-12-07
Do not show results panel when dry running

yuumasato:
When users starts a dry run scan they should only see the dialog with
the command that would be performed.

RHBZ#1385549

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -1155,6 +1155,12 @@ void MainWindow::scanCanceled()
 void MainWindow::scanFinished()
 {
     scanEnded(false);
+
+    // Clean results if the scan is a dry run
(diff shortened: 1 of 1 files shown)


src/MainWindow.cpp
```diff
@@ -1155,6 +1155,12 @@ void MainWindow::scanCanceled()
 void MainWindow::scanFinished()
 {
     scanEnded(false);
+
+    // Clean results if the scan is a dry run
+    // User will see the CommandLineArgsDialog and then the MainWindow
+    // ready for a new scan, or dry run
+    if (mUI.dryRunCheckBox->isChecked())
+        clearResults();
```
rsprudencio: The dry-run is actually ran and then results are cleared? Then it's not a dry-run, is it?

PR #94, opened 2016-12-07
Fix scan capabilities query

yuumasato:
Check if oscap is available in remote machine before checking capabilities.

RHBZ#1368463

### DIFF
--- a/src/OscapScannerRemoteSsh.cpp
+++ b/src/OscapScannerRemoteSsh.cpp
@@ -107,10 +107,28 @@ void OscapScannerRemoteSsh::evaluate()
         return;
     }
 
-    emit infoMessage(QObject::tr("Querying capabilities on remote machine..."));
-
     {
         SshSyncProcess proc(mSshConnection, this);
+        emit infoMessage(QObject::tr("Checking if oscap is available on remote machine..."));
+
+        proc.setCommand(QString("hash"));
+        proc.setArguments(QStringList(SCAP_WORKBENCH_REMOTE_OSCAP_PATH));
+        proc.setCancelRequestSource(&mCancelRequested);
+        proc.run();
+
+        if (proc.getExitCode() != 0)
+        {
+            emit errorMessage(
+                QObject::tr("Failed to locate oscap on remote machine. "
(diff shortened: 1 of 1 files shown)


src/OscapScannerRemoteSsh.cpp
```diff
@@ -107,10 +107,28 @@ void OscapScannerRemoteSsh::evaluate()
         return;
     }
 
-    emit infoMessage(QObject::tr("Querying capabilities on remote machine..."));
-
     {
         SshSyncProcess proc(mSshConnection, this);
+        emit infoMessage(QObject::tr("Checking if oscap is available on remote machine..."));
+
+        proc.setCommand(QString("hash"));
```
mpreisler: This relies on /bin/bash being the interpreter for the user. We may want to go with something more portable. Some people use zsh or even just sh.

src/OscapScannerRemoteSsh.cpp
```diff
@@ -107,10 +107,28 @@ void OscapScannerRemoteSsh::evaluate()
...
+        proc.setArguments(QStringList(SCAP_WORKBENCH_REMOTE_OSCAP_PATH));
+        proc.setCancelRequestSource(&mCancelRequested);
+        proc.run();
+
+        if (proc.getExitCode() != 0)
+        {
+            emit errorMessage(
+                QObject::tr("Failed to locate oscap on remote machine. "
+                        "Please, check that openscap-scanner is installed on remote machine.")
```
mpreisler: Add "the": installed on the remote machine.

src/OscapScannerRemoteSsh.cpp
yuumasato: Would `command -v` be a better option?

mpreisler: I think people mostly use `which` when they want to be compatible. It's in coreutils.

mpreisler: I will look into whether `command -v` is portable, I don't have a clue.

mpreisler:
It seems to be portable - http://pubs.opengroup.org/onlinepubs/009696899/utilities/command.html

PR #95, opened 2016-12-07
Check capabilities of oscap with propper arguments

yuumasato:
Check capabilities of oscap by checking its version with
command `oscap -V`


mpreisler:
ACK

PR #97, opened 2016-12-14
Fix --fetch-remote-resources warning with older openscap

yuumasato:
These changes makes workbench handle nicely a --fetch-remote-resources
warning message received through stdout.

This fix the use case when SCAP Workbench is running a scan in a machine 
with openscap prior to version 1.2.11.

This also removes the dependency over OpenSCAP 1.2.11 at compile time.


mpreisler:
Looks good.

PR #98, opened 2016-12-19
Show number of selected rules in profile listing

### LINKED ISSUE #40: [RFE] Hide 'default' profile if it does not contain some selected rules

yuumasato:
Show number of selected rules for each profile.
This change may avoid confusion on the semantics of the default profile.
Fixes #40 

Part of this PR depends on changes made to OpenSCAP, see https://github.com/OpenSCAP/openscap/pull/607.

So, this PR also adds checks for the OpenSCAP version we are linking with. 
If workbench is compiled with OpenSCAP up to version 1.2.12, we don't count the selected rules for default profiles.


mpreisler:
I have a few nitpicks about variable scoping and stuff like that but I will fix that myself to save time before the release. There is no need to declare variables at the beginnings of scopes in C++ or even in modern C and it's harmful to do it. The code is less readable and more error prone. The smaller the scope of a variable the better.

But ACK.

PR #99, opened 2016-12-19
fix manual compilation in subdirectory

finkr:
In Debian, compiling package for multi arch occur in directory.
doc/user_manual.html compilation fails with: `No rule to make target '../doc/user_manual.html', needed by 'CMakeFiles/docs'.  Stop.`

So using full path for CMakeLists's DEPENDS, as suggested here https://public.kitware.com/Bug/view.php?id=10217

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -214,7 +214,7 @@ install(FILES "scap-workbench.appdata.xml"
 if (ASCIIDOCTOR_EXECUTABLE)
     file(GLOB USER_MANUAL_SCREENSHOTS "${CMAKE_CURRENT_SOURCE_DIR}/doc/user_manual/*.png")
     add_custom_command(
-        OUTPUT doc/user_manual.html
+        OUTPUT ${CMAKE_CURRENT_SOURCE_DIR}/doc/user_manual.html
         COMMAND ${CMAKE_COMMAND} -E make_directory ${CMAKE_CURRENT_BINARY_DIR}/doc
         COMMAND ${ASCIIDOCTOR_EXECUTABLE} --destination-dir ${CMAKE_CURRENT_BINARY_DIR}/doc -b html5 -a data-uri user_manual.adoc
         COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_CURRENT_BINARY_DIR}/doc/user_manual.html ${CMAKE_CURRENT_SOURCE_DIR}/doc/user_manual.html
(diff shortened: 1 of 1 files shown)


yuumasato:
@openscap-jenkins test this please

yuumasato:
@openscap-jenkins ok to test

rsprudencio:
@openscap-jenkins add to white list

rsprudencio:
@openscap-jenkins test this please

CMakeLists.txt
```diff
@@ -214,7 +214,7 @@ install(FILES "scap-workbench.appdata.xml"
 if (ASCIIDOCTOR_EXECUTABLE)
     file(GLOB USER_MANUAL_SCREENSHOTS "${CMAKE_CURRENT_SOURCE_DIR}/doc/user_manual/*.png")
     add_custom_command(
-        OUTPUT doc/user_manual.html
+        OUTPUT ${CMAKE_CURRENT_SOURCE_DIR}/doc/user_manual.html
```
mpreisler: You meant `CURRENT_BINARY_DIR`, right? See:
```
COMMAND ${ASCIIDOCTOR_EXECUTABLE} --destination-dir ${CMAKE_CURRENT_BINARY_DIR}/doc -b html5 -a data-uri user_manual.adoc
```

finkr: Well, yes and no.

On one side, it seemed reasonable to produce the output in the BINARY directory, but...

The `install`line is installing from `${CMAKE_CURRENT_SOURCE_DIR}` :

    install(FILES "${CMAKE_CURRENT_SOURCE_DIR}/doc/user_manual.html"

... so I assumed it made sense that `add_custom_command`'s `OUTPUT` should also point there (??)

yuumasato: Why do we copy the user manual from `${CMAKE_CURRENT_BINARY_DIR}/doc/` to `${CMAKE_CURRENT_SOURCE_DIR}/doc/` (line 220) just to install from `${CMAKE_CURRENT_SOURCE_DIR}/doc`?

I think it makes sense to output to, and install from `CMAKE_CURRENT_BINARY_DIR`.

mpreisler: There is a very good reason why we  do that. It's so that when you use the tarball you don't have to have `asciidoctor` installed. `asciidoctor` is awesome but not packaged on all our target distributions, including RHEL.

mpreisler:
@finkr Are you using the release tarball? It has `user_manual.html` prebuilt in `docs/`.

finkr:
@mpreisler Debian packaging best practice encourage to rebuild everything from source (so anything can be modified/patch/forked...). So even though I grab the tarball, I actually force recompiling the documentation (by removing it before building).

Regarding the variable to use, I understand your point _build in binary directory_.
I tested it on my debian package and both work. I'll update my pull request.

mpreisler:
@finkr This breaks the build for me in case I use the tarball and don't have asciidoctor on the machine. That is one of our critical use-cases because RHEL doesn't have asciidoctor.

mpreisler:
Closed for inactivity.

PR #107, opened 2017-01-17
Comment why keep colon when reading download file

yuumasato:
Better document why we keep colon when reading stdout from openscap
and reading status is RS_READING_DOWNLOAD_FILE.


mpreisler:
ack

Issue #36, opened 2015-08-18
[RFE / bug] Automatically enable parent XCCDF group in the tailoring window, if the child node is selected (when "Deselect all" option was used before)


iankko:
scap-workbench (version 1.1.0) in the tailoring window does not automatically enables parent XCCDF group in the benchmark as selected, if child XCCDF rule (node) is selected.

Reproducible:
Always

Affected versions:
1.1.0 and below (didn't try current master)

How to reproduce:
1) Open some SCAP content in scap-workbench (will describe the process on SSG content for Fedora)
2) Select 'common' profile && click customize. Click 'Ok' on the next dialog displayed, this should take you to the main tailoring dialog,
3) Click "Deselect all" button / option && see how all the rules are deselected,
4) Now select some unselected group, check in the check-box next to it, and select also some child rule (deeper than one level in the tree). Check in the check-box next to that rule.
5) Click 'Confirm Changes' button

Actual result:
No rule is selected in the scap-workbench main window.

Expected result:
The selected child node / rule should be displayed in the main scap-workbench window.

Additional information:
Concrete scenario for 'Disable snmpd Service rule' is below:
1) Open 'Common Profile for General-Purpose Fedora Systems' Fedora SSG profile in scap-workbench,
2) Click 'Customize' button,
3) Click 'OK' when prompted for new profile ID etc.,
4, Click 'Deselect All' button,
5, Go to 'Services', enable 'SNMP Server' tree part, enable the 'Disable snmpd Service' rule too
6, Click 'Confirm changes'

Actual result:
Watch  how XCCDF Subgroup 'Disable SNMP Server if Possible' is group is not selected, and therefore 'Disable snmpd Service' rule is not selected either, and after clicking 'Confirm changes' there's no rule visible in scap-workbench's main window.

Expected result:
Selecting / checking in a main XCCDF group, and a rule within it, should also automatically checking all the parent groups, that are necessary in order to the final child rule to be included into the tailoring file (once Deselect All option has been used).

Not sure if this is bugfix / RFE, thus used both of them in the report.

Thank you for looking into this!

Jan

iankko:
[...]

Issue #23, opened 2015-06-29
Deselect All does not deselect all rules


zemb:
During tailoring, the rules that were checked before tailoring stay intact when I click Deselect All.

Only those that I check first, get deselected this way. Is this the way it is supposed to work?

scap-workbench 1.0.2, compiled with Qt 4.8.5, using openscap 1.2.4

ybznek:
@zemb I cannot reproduce your issue and I don't see any changes in the code what could fix your problem after you've filled it.

I can see, that Groups are not deselected during "deselect all", but I am not sure if it is a bug.

Can you still reproduce it?

yuumasato:
Fixed by #113

PR #113, opened 2017-05-05
Implements Tri-State checkboxes on Tailoring screen.

### LINKED ISSUE #112: Tailoring Screen: Checkboxes stay disabled after deselect all
Some checkboxes stay disabled after deselect all even if their parents and grandparents are active. To enable them again it's necessary to disable and re-enable their direct parents.

Steps:
step1
step2
step3
step4

Results:
result

WesleyCeraso:
Closes #112

### DIFF
--- a/src/TailoringWindow.cpp
+++ b/src/TailoringWindow.cpp
@@ -35,6 +35,7 @@
 #include <QUndoView>
 #include <QTimer>
 #include <QDateTime>
+#include <QStack>
 
 #include <algorithm>
 #include <cassert>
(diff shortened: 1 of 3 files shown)


WesleyCeraso:
Also fixes #36, #23

WesleyCeraso:
@openscap-jenkins retest this please

rsprudencio:
@openscap-jenkins retest this please

mpreisler:
This is pretty cool and seems to be working well. It is however very slow. Could we make it faster? I suspect it's probably the repeated wasted redraws that's making it slow.

If we can't make it faster we should at least show a busy cursor to let the user know we are doing something and the app hasn't crashed.

mpreisler:
@yuumasato I would appreciate your review here as well. This is a big change.

WesleyCeraso:
@mpreisler I  haven't noticed how slow it was when painting big trees. In fact it was really slow when a big part of the tree is expanded and you check/uncheck the most top item. It's now fixed.

src/TailoringWindow.cpp
```diff
@@ -268,6 +268,7 @@ void TailoringWindow::synchronizeProfileItem()
 void TailoringWindow::synchronizeTreeItem()
 {
     synchronizeTreeItemSelections(mBenchmarkItem);
+    return;
```
mpreisler: ?

src/TailoringWindow.cpp
```diff
@@ -315,6 +322,7 @@ void TailoringWindow::synchronizeTreeItemSelections(QTreeWidgetItem* treeItem)
                     treeItem->setCheckState(0, Qt::Unchecked);
                 }
                 break;
+
```
mpreisler: ?

mpreisler:
It's faster now but when I select/deselect it screws up the scroll position and scrolls all the way up.

mpreisler:
It's much better to just disable repainting while you change the tree. That's easier than collapsing and expanding parts of the tree.

WesleyCeraso:
@mpreisler Yeah, I'll take take a better look at this.

WesleyCeraso:
@mpreisler The problem was not exactly with the painting event, as it can be postponed by setUpdatesEnabled(bool enable), instead, if the treewidget was visible, for each data changed it would recalculate its geometry.
The only way I found to prevent it is turning the treewidget invisible. The last solution I gave (collapsing) did the same thing but would lost the scroll position as you noticed.

src/TailoringWindow.cpp
```diff
@@ -300,160 +265,71 @@ void TailoringWindow::synchronizeProfileItem()
...
 
-void TailoringWindow::synchronizeTreeItem(QTreeWidgetItem* treeItem, struct xccdf_item* xccdfItem, bool recursive)
+void TailoringWindow::synchronizeTreeItem()
 {
-    ++mSynchronizeItemLock;
+    // setting visible as false seems to be the only way to avoid entering the "if" in
+    // https://github.com/qt/qt/blob/4.8/src/gui/itemviews/qabstractitemview.cpp#L3190
+    mUI.itemsTree->setVisible(false);
+    synchronizeTreeItemSelections(mBenchmarkItem);
+    mUI.itemsTree->setVisible(true);
```
yuumasato: Please, make it clearer that we are making it invisible and visible again due to perceived slowness when changing state of long trees.

WesleyCeraso:
@yuumasato

WesleyCeraso:
@yuumasato Empty groups are checked by default so that their parent can be checked, instead of partially checked, in case all its siblings are checked as well.

yuumasato:
@WesleyCeraso That makes sense. 
These are great changes, thank you!
Ack.

PR #114, opened 2017-05-11
Fix garbled XHTML formatted profile description

### LINKED ISSUE #91: XHTML formatted profile description shows up garbled
image

Could be as simple as changing the type of the widget, needs investigation.

WesleyCeraso:
Fixes #91

### DIFF
--- a/src/APIHelpers.cpp
+++ b/src/APIHelpers.cpp
@@ -31,11 +31,9 @@ extern "C"
 
 QString oscapTextIteratorGetPreferred(struct oscap_text_iterator* it, const QString& lang)
 {
-    char* preferred_s = oscap_textlist_get_preferred_plaintext(it, lang.isEmpty() ? NULL : lang.toUtf8().constData());
+    oscap_text* preferred_s = oscap_textlist_get_preferred_text(it, lang.isEmpty() ? NULL : lang.toUtf8().constData());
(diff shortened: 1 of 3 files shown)


WesleyCeraso:
Hi Team,

I don't know if there is a clever way to remove the html namespace from the description. If there is, please, let me know.

rsprudencio:
Please use description as well on commits so we don't get lost using only ticket numbers.

rsprudencio:
LGTM :smile:

mpreisler:
Have you checked whether this is exposed in all versions of OpenSCAP that we support in this branch? This function used to be private and I think that's why it's not used in SCAP Workbench.

```
$ git show b64883fa
commit b64883fa65738bd4f9b81fd8584a778d8d1957f4
Author: Martin Preisler <mpreisle@redhat.com>
Date:   Wed Nov 5 11:18:27 2014 +0100

    Expose oscap_textlist_get_preferred_text in public API
    
    This function is needed to get XHTML descriptions for scap-workbench.
```

WesleyCeraso:
@mpreisler I relied on this function because there is already another, 3 years old, use of it in https://github.com/OpenSCAP/scap-workbench/blob/v1-1/src/APIHelpers.cpp#L57

PR #116, opened 2017-05-30
Reads "Downloading ..." message from stderr

WesleyCeraso:
Oscap used to send this message through stdout. When it changed to stderr workbench was not updated.

### DIFF
--- a/src/OscapScannerBase.cpp
+++ b/src/OscapScannerBase.cpp
@@ -306,6 +306,7 @@ bool OscapScannerBase::tryToReadStdOutChar(QProcess& process)
         {
             case RS_READING_PREFIX:
                 {
+                    // Openscap <= 1.2.10 (60fb9f0c98eee) sends this message through stdout
                     if (mReadBuffer=="Downloading")
                     {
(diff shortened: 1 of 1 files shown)


yuumasato:
I wish there were comments in source explaining why we parse the `Downloading` message in both stdout and stderr.

WesleyCeraso:
@yuumasato You can find the comments now.

src/OscapScannerBase.cpp
```diff
@@ -435,6 +436,11 @@ void OscapScannerBase::watchStdErr(QProcess& process)
                 QString guiMessage = guiFriendlyMessage(stdErrOutput);
                 emit warningMessage(QObject::tr(guiMessage.toUtf8().constData()));
             }
+            // Openscap >= 1.2.11 (60fb9f0c98eee) sends this message through stderr
+            if (stdErrOutput.contains(QRegExp("^Downloading: .+ \\.{3} \\w+\\n")))
```
mpreisler: should be `else if`

WesleyCeraso: Sure

PR #117, opened 2017-06-04
MainWindow tab order and SSG Dialog default button

WesleyCeraso:
These two commits fixes the tab order in MainWindow and the default button in SSG Integration Dialog respectively.

### DIFF
--- a/src/RemoteMachineComboBox.cpp
+++ b/src/RemoteMachineComboBox.cpp
@@ -26,6 +26,7 @@ RemoteMachineComboBox::RemoteMachineComboBox(QWidget* parent):
     QWidget(parent)
 {
     mUI.setupUi(this);
+    setFocusProxy(mUI.host);
 
 #if (QT_VERSION >= QT_VERSION_CHECK(4, 7, 0))
     // placeholder text is only supported in Qt 4.7 onwards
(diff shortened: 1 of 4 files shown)


yuumasato:
I noticed that the Rules list is tabbed after the `Scan` button, can it be placed after the `Expand all` button?

Also, something unrelated that I noticed, pressing <Enter> on `Scan` button does nothing, you need to use <Space> to activate it, do you know why?

WesleyCeraso:
@yuumasato The Rules list is in the correct tab order now. About the Scan button I could not reproduce what you described, instead, it always start the scan here.

yuumasato:
@WesleyCeraso Cool. About the `Scan` button, I just noticed that I used weird formatting and github just hid the buttons I mentioned.

WesleyCeraso:
@yuumasato Yes, I know why. When you want to "click" whatever has focus you should press Space. When you press Enter, the default button will be "clicked", howeverm there is no default button in the MainWindow.

I'll make te Scan button the default button.

WesleyCeraso:
Never mind, default button only works inside Dialogs, not MainWindows. There are workarounds, but I don't think they are worthwhile.
That said, you should expect to use the space, not enter.

yuumasato:
> When you want to "click" whatever has focus you should press Space. When you press Enter, the default button will be "clicked",

Interesting.

> Never mind, default button only works inside Dialogs, not MainWindows. There are workarounds, but I don't think they are worthwhile.
That said, you should expect to use the space, not enter.

Ok.

Thank you for these fixes!

PR #119, opened 2017-06-06
Validates custom profile id on-the-fly

### LINKED ISSUE #49: Non intuitive "profile ID" text edit.
When you want to customize some profile, you can type name of new profile.
If you remove "xccdf_" prefix part, text edit doesn't allow you to type(without any notification), if you don't type "xccdf_"

I think, it would be better if text edit will contain "xccdf_" prefix always and doesn't allow to delete it.

In my opinion, "OK" button should not be enabled if you don't have valid ID. Message when dialog is closed is not user friendly.

WesleyCeraso:
Closes #49

### DIFF
--- a/include/TailorProfileDialog.h
+++ b/include/TailorProfileDialog.h
@@ -24,6 +24,7 @@
 
 #include "ForwardDecls.h"
 #include <QDialog>
+#include <QPushButton>
 
 #include "ui_TailorProfileDialog.h"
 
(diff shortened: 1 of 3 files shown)


WesleyCeraso:
I removed the QRegExpValidator because the way it behaves is not so intuitive and can be annoying. Instead, the code now only enables the OK button when the regex is matched.

yuumasato:
Ack.

PR #120, opened 2017-06-06
Refactors "Deselect all" and adds Group Id to Undo Message

WesleyCeraso:


src/TailoringWindow.cpp
```diff
@@ -40,6 +40,10 @@
 #include <algorithm>
 #include <cassert>
 
+const QString DESELECT_ALL("Deselect All");
+const QString GROUP_ACTIVATION("Group %1 activated");
+const QString GROUP_DEACTIVATION("Group %1 deactivated");
```
yuumasato: Is the proper wording activated/deactivated or selected/deselected?

yuumasato:
Also, selection and deselection message of Groups and Rules in Undo stack are not consistent.
And I think that showing title (like the rule messages) is much easier for user than showing ID.

WesleyCeraso:
Hi @yuumasato, I checked the string for items and they were "select "%1"" and "unselect "%2"". However, I searched for "unselect" and it does not seems to be a valid word, so I changed to "deselect".

In order to improve consistence, now we have:
'select "%1"' and 'deselect "%1"' for items
'select group "%1"' and 'deselect group "%1"' for groups
'deselect all'

where %1 means the title.

WesleyCeraso:
Now the "Deselect All" button gets disabled when there is no item checked.

yuumasato:
It looks awesome! 
I have a few nitpicks, all items are actually rules, right? So it would be interesting to add "Rule" to the message.
And I think the messages look better capitalized, the same for the words "Rule" and "Group".

WesleyCeraso:
@yuumasato Updated.

I also gave up on the macro-like strings and used the text directly.

src/TailoringUndoCommands.cpp
```diff
@@ -148,9 +148,9 @@ void XCCDFItemSelectUndoCommand::refreshText()
         title = QString::fromUtf8(xccdf_item_get_id(xccdfItem));
 
     if (mNewSelect)
-        setText(QObject::tr("select \"%1\"").arg(title));
+        setText(QObject::tr("Select rule \"%1\"").arg(title));
     else
-        setText(QObject::tr("unselect \"%1\"").arg(title));
+        setText(QObject::tr("Deselect rule \"%1\"").arg(title));
```
yuumasato: Please make "rule" capitalized.

src/TailoringWindow.cpp
```diff
@@ -965,33 +953,15 @@ void TailoringWindow::itemChanged(QTreeWidgetItem* treeItem, int column)
...
-                mUndoStack.push(new XCCDFItemSelectUndoCommand(this, curr, checkState));
-            }
-
-            for (int i = 0; i < curr->childCount(); ++i)
-                stack.push(curr->child(i));
-        }
-
-        mUndoStack.endMacro();
+        const QString title = oscapItemGetReadableTitle(xccdfItem, mPolicy);
+        createSelectionMacro(treeItem, checkState, checkState ? QObject::tr("Select group \"%1\"").arg(title) : QObject::tr("Deselect group \"%1\"").arg(title));
```
yuumasato: The same with "group", make it "Group"

yuumasato:
Thank you, Wesley. Ack.

PR #121, opened 2017-06-08
Fixes the focus of remote machine section

WesleyCeraso:
Because of the TabFocus, clicks were not reaching the "User and host" field.


yuumasato:
Ack.

PR #122, opened 2017-06-09
Fixes Cancel process

### LINKED ISSUE #71: Interrupted remote scan - active progressbar/"processing" remain
I have interrupted scan and "The oscap process didn't terminate in time, it will be killed instead"

screenshot_2016-07-13_12-29-49

But it seems that scan is still working (according to progressbar and "processing" text)

WesleyCeraso:
The cancel process kills the top most pkexec script, the oscap script notices it and try to send a SIGINT to the oscap tool, however, because of the SIGNAL trap and because SIGINT would never reach the oscap tool (even without the trap) everything keeps running.

Reference from bash manpage:
Non-builtin commands run by bash have signal handlers set to the values inherited by the shell from its parent. When job control is not in effect, asynchronous commands ignore SIGINT and SIGQUIT in addition to these inherited handlers.

Fixes #71

### DIFF
--- a/scap-workbench-oscap.sh
+++ b/scap-workbench-oscap.sh
@@ -18,7 +18,7 @@
 
 set -u -o pipefail
 
-trap "" SIGHUP SIGINT SIGTERM
+trap "" SIGHUP SIGINT
 
 # pkexec writes a message to stderr when user dismisses it, we always skip 1 line.
 # if user did not dismiss it we should print a dummy line to stderr so that nothing
@@ -74,7 +74,7 @@ while kill -0 $PID 2> /dev/null; do
     ret=$?
     if [ 0 -lt $ret -a $ret -lt 128 ]; then
         # If read failed & it was not due to timeout --> parents are gone.
-        kill -s SIGINT $PID 2> /dev/null
+        kill -s SIGTERM $PID 2> /dev/null
         break
     fi
 done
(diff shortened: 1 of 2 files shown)


scap-workbench-oscap.sh
```diff
@@ -74,7 +74,7 @@ while kill -0 $PID 2> /dev/null; do
     ret=$?
     if [ 0 -lt $ret -a $ret -lt 128 ]; then
         # If read failed & it was not due to timeout --> parents are gone.
-        kill -s SIGINT $PID 2> /dev/null
+        kill -TERM $PID 2> /dev/null
```
mpreisler: We shouldn't just TERM the process, that will leave dangling resources. Is "TERM" really needed there?

WesleyCeraso: Well, I'm really not an expert on which signal to use here, however, SIGINT is not an option as it does not reach the oscap tool.

What signal do you think that fits better?

mpreisler: I looked at it again and you are right, we should use TERM here.

However please change the `kill` syntax to `kill -s SIGTERM $PID`. According to the manpage that's more portable.

WesleyCeraso: Updated

mpreisler:
ACK

PR #124, opened 2017-06-13
Capitalizes "Set Value" undo message

WesleyCeraso:
Capitalizes "Set Value" undo message to make it consistent with other messages


yuumasato:
Ack.

PR #127, opened 2017-06-14
Fixes memory leak as TailoringScreen is never deleted

WesleyCeraso:


mpreisler:
ACK

PR #130, opened 2017-06-21
Fixes the long time to open TailoringWindow in MacOS

WesleyCeraso:
There seems to be a bug in Qt (MacOS implementation) when it tries to calculate geometry/draw tri-state checkboxes that makes it spend much more time than usual.

This commit changes the traversal of the tree from preorder to postorder so that we set the items as expanded while they are still invisible, thus avoiding the geometry calculation and drawing. Only when the top-most item is expanded that all items are drawed, once.

### DIFF
--- a/src/TailoringWindow.cpp
+++ b/src/TailoringWindow.cpp
@@ -664,6 +664,9 @@ void TailoringWindow::syncCollapsedItem(QTreeWidgetItem* item, QSet<QString>& us
 {
     struct xccdf_item* xccdfItem = getXccdfItemFromTreeItem(item);
     const QString id = QString::fromUtf8(xccdf_item_get_id(xccdfItem));
+    
+    for (int i = 0; i < item->childCount(); ++i)
+        syncCollapsedItem(item->child(i), usedCollapsedIds);
 
     if (mCollapsedItemIds.contains(id))
     {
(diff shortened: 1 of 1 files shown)


mpreisler:
> There seems to be a bug in Qt (MacOS implementation) when it tries to calculate geometry/draw tri-state checkboxes that makes it spend much more time than usual.

Have you reported this to Qt? Is it really a bug or is this a case of us abusing the API?

WesleyCeraso:
@mpreisler Seems to be related to https://bugreports.qt.io/browse/QTBUG-9271, and it is there since 2010. I don`t think they will fix it anymore.

mpreisler:
> @mpreisler It seems to be related to https://bugreports.qt.io/browse/QTBUG-9271, and it is there since 2010. I don`t think they will fix it anymore.

That is related to QWindowsStyle which we aren't using, so I don't think it's the same bug.

WesleyCeraso:
@mpreisler But QMacStyle inherits QWindowsStyle (strange huh?). I don't think it's worthwhile to dig deeper to determine if it is or not the same bug. It's seems clear it's Qt's fault in MacOs implementation because you can execute the same test on different OSs and they perform well, even though we are abusing it in all platforms.

mpreisler:
That is strange.

Very well, ACK.

PR #131, opened 2017-06-26
Fixes the focus in tailoring window when (un)checking checkboxes

### LINKED ISSUE #128: Checking checkboxes in the tailoring window temporarily loses focus on the tree widget
You can observe the tree widget changing colors from black to grey and back as it loses focus every time you click on anything. This is probably easy to fix by manipulating updates and redraws while changing the tree.

WesleyCeraso:
Fixes #128

### DIFF
--- a/src/TailoringWindow.cpp
+++ b/src/TailoringWindow.cpp
@@ -264,12 +264,19 @@ void TailoringWindow::synchronizeProfileItem()
 
 void TailoringWindow::synchronizeTreeItem()
 {
+    // If itemsTree has focus, it will lose it when it becomes invisible
+    QWidget* focused = QApplication::focusWidget();
+    
     // If QTreeWidget remains visible during the sync, it'll recalculate its geometry for each checked
     // row. Configuring "visible" as false seems to be the only way to avoid entering the "if" in
     // https://github.com/qt/qt/blob/4.8/src/gui/itemviews/qabstractitemview.cpp#L3190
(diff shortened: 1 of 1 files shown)


src/TailoringWindow.cpp
```diff
@@ -267,9 +267,12 @@ void TailoringWindow::synchronizeTreeItem()
     // If QTreeWidget remains visible during the sync, it'll recalculate its geometry for each checked
     // row. Configuring "visible" as false seems to be the only way to avoid entering the "if" in
     // https://github.com/qt/qt/blob/4.8/src/gui/itemviews/qabstractitemview.cpp#L3190
+    QWidget * focused = QApplication::focusWidget();
```
mpreisler: coding style

src/TailoringWindow.cpp
```diff
@@ -267,9 +267,12 @@ void TailoringWindow::synchronizeTreeItem()
...
     mUI.itemsTree->setVisible(false);
     synchronizeTreeItemSelections(mBenchmarkItem);
     mUI.itemsTree->setVisible(true);
+    // The widget loses focus when it becomes invisible, so we set it again
+    if (focused) focused->setFocus();
```
mpreisler: coding style

src/TailoringWindow.cpp
mpreisler: I also think a comment should explain why we are doing all this.

PR #142, opened 2017-09-22
Port needs to be unsigned short, otherwise we run risks of overflow

### LINKED ISSUE #140: scap-workbench remote machine with (example port) 50.000 or 60.000, result port 22 in dry run.
scap-workbench remote machine with port 50.000 or 60.000, result port 22 in dry run.

Produced in:
CentOS 7 x64 or RHEL 7 x64
SCAP Workbench 1.1.4, compiled with Qt 4.8.5, using OpenSCAP 1.2.14

Open it, click remote machine, change port to 50000 or 60215 or 60000
dry run
click scan
result: oscap-ssh root@target-ip 22 ...............................................................

I see at maximum it correctly display until 32767.

mpreisler:
60000 is a valid port but can't be represented via just "short".

Fixes #140

### DIFF
--- a/include/OscapScannerRemoteSsh.h
+++ b/include/OscapScannerRemoteSsh.h
@@ -31,7 +31,7 @@ class OscapScannerRemoteSsh : public OscapScannerBase
     Q_OBJECT
 
     public:
-        static void splitTarget(const QString& in, QString& target, short& port);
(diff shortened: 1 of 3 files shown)


WesleyCeraso:
ACK

PR #143, opened 2017-10-11
Generate bash and ansible remediation roles from profiles

matejak:
Before this PR, S-W was able to remediate on the fly or remediate after the check using `oscap`.
This PR aims to enable users to control the remediation by giving them access to remediation scripts (that they can edit and execute by themselves).

EDIT by @mpreisler: Let's only do generate from profiles here and generate from results can be in a separate PR.

### DIFF
--- a/include/MainWindow.h
+++ b/include/MainWindow.h
@@ -420,9 +420,9 @@ class MainWindow : public QMainWindow
         void toggleRuleResultsExpanded();
         void toggleRuleResultsExpanded(bool checked);
 
-        /// Pops up a save dialog for a bash remediation
+        /// Pops up a save dialog for a bash remediation (based just on the currently selected profile)
         void generateBashRemediationRole();
-        /// Pops up a save dialog for an ansible remediation
+        /// Pops up a save dialog for an ansible remediation (based just on the currently selected profile)
         void generateAnsibleRemediationRole();
 
     public slots:
--- a/include/MainWindow.h
+++ b/include/MainWindow.h
@@ -424,6 +424,8 @@ class MainWindow : public QMainWindow
         void generateBashRemediationRole();
         /// Pops up a save dialog for an ansible remediation (based just on the currently selected profile)
         void generateAnsibleRemediationRole();
+        /// Pops up a save dialog for an puppet remediation (based just on the currently selected profile)
+        void generatePuppetRemediationRole();
 
     public slots:
         /**
--- a/include/MainWindow.h
+++ b/include/MainWindow.h
@@ -421,9 +421,9 @@ class MainWindow : public QMainWindow
         void toggleRuleResultsExpanded(bool checked);
 
         /// Pops up a save dialog for a bash remediation
-        void genBash();
+        void generateBashRemediationRole();
         /// Pops up a save dialog for an ansible remediation
-        void genAnsible();
+        void generateAnsibleRemediationRole();
 
     public slots:
         /**
--- a/include/MainWindow.h
+++ b/include/MainWindow.h
@@ -420,6 +420,11 @@ class MainWindow : public QMainWindow
         void toggleRuleResultsExpanded();
         void toggleRuleResultsExpanded(bool checked);
 
+        /// Pops up a save dialog for a bash remediation
+        void genBash();
+        /// Pops up a save dialog for an ansible remediation
+        void genAnsible();
+
     public slots:
         /**
          * @brief Changes state of Expand all/Collapse all button according to boolean received
--- a/include/RemediationRoleSaver.h
+++ b/include/RemediationRoleSaver.h
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013 Red Hat Inc., Durham, North Carolina.
+ * Copyright 2017 Red Hat Inc., Durham, North Carolina.
  * All Rights Reserved.
  *
  * This program is free software: you can redistribute it and/or modify
@@ -16,7 +16,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  *
  * Authors:
- *      Martin Preisler <mpreisle@redhat.com>
+ *      Matej Tyc <matyc@redhat.com>
  */
 
 #ifndef SCAP_WORKBENCH_REMEDIATION_ROLE_SAVER_H_
@@ -40,8 +40,9 @@ extern "C"
 class RemediationSaverBase
 {
     public:
-    RemediationSaverBase(QWidget* parentWindow, ScanningSession* session): mParentWindow(parentWindow), mScanningSession(session) {}
-        void SelectFilenameAndSaveRole();
+        RemediationSaverBase(QWidget* parentWindow, ScanningSession* session):
+            mParentWindow(parentWindow), mScanningSession(session) {}
+        void selectFilenameAndSaveRole();
 
     protected:
         const ScanningSession* mScanningSession;
@@ -53,8 +54,8 @@ class RemediationSaverBase
         QString mFixTemplate;
 
     private:
-        int SaveToFile(const QString& filename);
-        QString guessFilenameStem();
+        int saveToFile(const QString& filename);
+        QString guessFilenameStem() const;
 };
 
 
--- a/include/RemediationRoleSaver.h
+++ b/include/RemediationRoleSaver.h
@@ -1,0 +1,65 @@
+#ifndef SCAP_WORKBENCH_REMEDIATION_ROLE_SAVER_H_
+#define SCAP_WORKBENCH_REMEDIATION_ROLE_SAVER_H_
+
+#include "ForwardDecls.h"
+#include "ScanningSession.h"
+
+#include <QString>
+#include <QFile>
+#include <QFileDialog>
+
+extern "C"
+{
+#include <xccdf_benchmark.h>
+#include <xccdf_policy.h>
+#include <xccdf_session.h>
+}
+
+
+class RemediationSaverBase
+{
+    public:
+	RemediationSaverBase(QWidget* parentWindow, ScanningSession* session): mParentWindow(parentWindow), mScanningSession(session) {}
+        void SelectFilenameAndSaveRole();
+
+    protected:
+        const ScanningSession* mScanningSession;
+        QWidget* mParentWindow;
+
+        QString mSaveMessage;
+        QString mFiletypeExtension;
+        QString mFiletypeTemplate;
+        QString mFixTemplate;
+
+    private:
+        int SaveToFile(const QString& filename);
+        QString guessFilenameStem();
+};
+
+
+class BashRemediationSaver : public RemediationSaverBase
+{
+    public:
+	BashRemediationSaver(QWidget* parentWindow, ScanningSession* session):RemediationSaverBase(parentWindow, session)
+	{
+	    mSaveMessage = QObject::tr("Save remediation role as a bash script");
+	    mFiletypeExtension = "sh";
+	    mFiletypeTemplate = QObject::tr("bash script (*.%1)");
+	    mFixTemplate = QObject::tr("urn:xccdf:fix:script:sh");
+        }
+};
+
+
+class AnsibleRemediationSaver : public RemediationSaverBase
+{
+    public:
+	AnsibleRemediationSaver(QWidget* parentWindow, ScanningSession* session):RemediationSaverBase(parentWindow, session)
+	{
+	    mSaveMessage = QObject::tr("Save remediation role as an ansible playbook");
+	    mFiletypeExtension = "yml";
+	    mFiletypeTemplate = QObject::tr("ansible playbook (*.%1)");
+	    mFixTemplate = QObject::tr("urn:xccdf:fix:script:ansible");
+        }
+};
+
+#endif // SCAP_WORKBENCH_REMEDIATION_ROLE_SAVER_H_
--- a/include/RemediationRoleSaver.h
+++ b/include/RemediationRoleSaver.h
@@ -1,3 +1,24 @@
+/*
+ * Copyright 2013 Red Hat Inc., Durham, North Carolina.
+ * All Rights Reserved.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authors:
+ *      Martin Preisler <mpreisle@redhat.com>
+ */
+
 #ifndef SCAP_WORKBENCH_REMEDIATION_ROLE_SAVER_H_
 #define SCAP_WORKBENCH_REMEDIATION_ROLE_SAVER_H_
 
@@ -19,7 +40,7 @@ extern "C"
 class RemediationSaverBase
 {
     public:
-	RemediationSaverBase(QWidget* parentWindow, ScanningSession* session): mParentWindow(parentWindow), mScanningSession(session) {}
+    RemediationSaverBase(QWidget* parentWindow, ScanningSession* session): mParentWindow(parentWindow), mScanningSession(session) {}
         void SelectFilenameAndSaveRole();
 
     protected:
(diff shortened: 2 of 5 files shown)


matejak:
The easier part has been done - you should be able to save remediations from the main app window.

include/MainWindow.h
```diff
@@ -420,6 +420,11 @@ class MainWindow : public QMainWindow
         void toggleRuleResultsExpanded();
         void toggleRuleResultsExpanded(bool checked);
 
+        /// Pops up a save dialog for a bash remediation
+        void genBash();
```
mpreisler: write a better method name please, it's worth it for self-documentation and editors auto-complete anyway.

mpreisler: also, can't we have just one method that gets "template" passed in? that way we can add even more remediation options easily later on.

src/MainWindow.cpp
```diff
@@ -1531,3 +1551,48 @@ QMessageBox::StandardButton MainWindow::openNewFileQuestionDialog(const QString&
...
+        QObject::tr("ansible playbooks (*.yml)"), 0
+#ifndef SCAP_WORKBENCH_USE_NATIVE_FILE_DIALOGS
+        , QFileDialog::DontUseNativeDialog
+#endif
+    );
+
+    if (filename.isEmpty())
+        return;
+
+    int output_fd = open(filename.toUtf8(), O_CREAT|O_TRUNC|O_NOFOLLOW|O_WRONLY, 0700);
```
mpreisler: if at all possible use Qt abstractions, AFAIK one of them gives you the `fd`.

mpreisler:
To avoid confusing users even more we have to stick to consistent terms. We use "remediation roles" in SCAP Security Guide and I think we should stick to it across all projects.

src/MainWindow.cpp
mpreisler: https://doc.qt.io/qt-4.8/qfile.html#handle

mpreisler:
@openscap-ci add to whitelist

include/MainWindow.h
matejak: I think that it is unavoidable to have multiple methods as one menu item requires one slot method. As noted in the code, there is a potential for code reuse between those methods. I am thinking of polymorphism and I would postpone this until the remediation from results is implemented, since there could be some code reuse with this approach as well.

include/MainWindow.h
```diff
@@ -421,9 +421,9 @@ class MainWindow : public QMainWindow
         void toggleRuleResultsExpanded(bool checked);
 
         /// Pops up a save dialog for a bash remediation
```
mpreisler: since we talk about rule results just above this, let's make sure the comment mentions that these roles are generated for the currently selected profile, not for results

src/MainWindow.cpp
```diff
@@ -247,6 +248,23 @@ MainWindow::MainWindow(QWidget* parent):
 
     // start centered
     move(QApplication::desktop()->screen()->rect().center() - rect().center());
+
```
mpreisler: should all of this really be done at the end of the ctor? Could you please improve the variable names? `mGenMenu` is not even a member and it starts with `m`.

src/MainWindow.cpp
```diff
@@ -1531,3 +1549,48 @@ QMessageBox::StandardButton MainWindow::openNewFileQuestionDialog(const QString&
           QMessageBox::Yes | QMessageBox::No, QMessageBox::No
     );
 }
+
+// TODO: Reuse code between generateBashRemediationRole and generateAnsibleRemediationRole
+void MainWindow::generateBashRemediationRole()
+{
```
mpreisler: refactor this into a method that takes the short name "bash remediation role", extension ".sh" and OpenSCAP remediation template "urn:xccdf:...". Then the two generate functions can just call that. Shorter code is better.

ui/MainWindow.ui
```diff
@@ -331,6 +358,16 @@
               </widget>
              </item>
              <item>
+              <widget class="QPushButton" name="genRemediationButton">
+               <property name="toolTip">
+                <string>Generate remediations of all possible failiures that the selected profile can cover.</string>
```
mpreisler: typos. again, please use consistent terms. It's a remediation role. It contains a fix for every single rule in the profile that has a fix.

ui/MainWindow.ui
```diff
@@ -331,6 +358,16 @@
...
+               </property>
+               <property name="text">
+                <string>Get profile remediations</string>
```
mpreisler: Generate remediation role?

src/MainWindow.cpp
```diff
@@ -1531,3 +1549,48 @@ QMessageBox::StandardButton MainWindow::openNewFileQuestionDialog(const QString&
...
+#endif
+    );
+
+    if (filename.isEmpty())
+        return;
+
+    QFile output_file(filename);
+    output_file.open(QIODevice::WriteOnly);
+    struct xccdf_session * session = mScanningSession->getXCCDFSession();
+    struct xccdf_policy *policy = xccdf_session_get_xccdf_policy(session);
```
mpreisler: code style for the above 2 lines. it sucks but code style in S-W is different from openscap

src/MainWindow.cpp
```diff
@@ -1531,3 +1549,48 @@ QMessageBox::StandardButton MainWindow::openNewFileQuestionDialog(const QString&
...
+    xccdf_policy_generate_fix(policy, NULL, "urn:xccdf:fix:script:sh", output_file.handle());
+    output_file.close();
+}
+
+void MainWindow::generateAnsibleRemediationRole()
+{
+    const QString filename = QFileDialog::getSaveFileName(this,
+        QObject::tr("Save as Ansible playbook"),
+        QObject::tr("remediate-profile.yml"),
+        QObject::tr("ansible playbooks (*.yml)"), 0
```
mpreisler: file format is singular: "ansible playbook", not "playbooks"

src/MainWindow.cpp
```diff
@@ -1531,3 +1549,48 @@ QMessageBox::StandardButton MainWindow::openNewFileQuestionDialog(const QString&
...
     );
 }
+
+// TODO: Reuse code between generateBashRemediationRole and generateAnsibleRemediationRole
+void MainWindow::generateBashRemediationRole()
+{
+    const QString filename = QFileDialog::getSaveFileName(this,
+        QObject::tr("Save as bash script"),
+        QObject::tr("remediate-profile.sh"),
+        QObject::tr("bash scripts (*.sh)"), 0
```
mpreisler: file format is singular: "bash script", not "scripts"

src/MainWindow.cpp
```diff
@@ -1531,3 +1549,48 @@ QMessageBox::StandardButton MainWindow::openNewFileQuestionDialog(const QString&
...
+    );
+
+    if (filename.isEmpty())
+        return;
+
+    QFile output_file(filename);
+    output_file.open(QIODevice::WriteOnly);
+    struct xccdf_session * session = mScanningSession->getXCCDFSession();
+    struct xccdf_policy *policy = xccdf_session_get_xccdf_policy(session);
+    xccdf_policy_generate_fix(policy, NULL, "urn:xccdf:fix:script:sh", output_file.handle());
```
mpreisler: error checking missing

src/MainWindow.cpp
```diff
@@ -1531,3 +1549,48 @@ QMessageBox::StandardButton MainWindow::openNewFileQuestionDialog(const QString&
...
+        QObject::tr("ansible playbooks (*.yml)"), 0
+#ifndef SCAP_WORKBENCH_USE_NATIVE_FILE_DIALOGS
+        , QFileDialog::DontUseNativeDialog
+#endif
+    );
+
+    if (filename.isEmpty())
+        return;
+
+    QFile output_file(filename);
```
mpreisler: code style (variable name)

src/MainWindow.cpp
```diff
@@ -1531,3 +1549,48 @@ QMessageBox::StandardButton MainWindow::openNewFileQuestionDialog(const QString&
           QMessageBox::Yes | QMessageBox::No, QMessageBox::No
     );
 }
+
+// TODO: Reuse code between generateBashRemediationRole and generateAnsibleRemediationRole
+void MainWindow::generateBashRemediationRole()
+{
+    const QString filename = QFileDialog::getSaveFileName(this,
+        QObject::tr("Save as bash script"),
+        QObject::tr("remediate-profile.sh"),
```
mpreisler: Could we please base the suggested file name on the benchmark and profile ID? that would increase UX because many times I can just click save without typing anything.

matejak: I keep this in mind, but I have postponed it. Do you have any idea what is a good form? I am not sure about benchmark IDs, but those profile IDs alone are quite long.

src/MainWindow.cpp
matejak: I took slightly different approach, so I hope you like it too as it has the same effect.

src/MainWindow.cpp
matejak: I basically repeated what I have seen in the `ResultsViewer` constructor. I don't know where else to put it. It seems that the pattern here is that the appearance of the form is fully determined after the ctor is executed, so from this POV, it makes sense to group the stuff that defines UI in there.

mpreisler:
The class doesn't follow code conventions and more importantly I think that's a sign of over-engineering. The class doesn't really hold or remember any state, it just runs something. In this case I think we should just use a function, it's going to be easier and shorter.

include/RemediationRoleSaver.h
```diff
@@ -0,0 +1,65 @@
...
+
+class BashRemediationSaver : public RemediationSaverBase
+{
+    public:
+	BashRemediationSaver(QWidget* parentWindow, ScanningSession* session):RemediationSaverBase(parentWindow, session)
+	{
+	    mSaveMessage = QObject::tr("Save remediation role as a bash script");
+	    mFiletypeExtension = "sh";
+	    mFiletypeTemplate = QObject::tr("bash script (*.%1)");
+	    mFixTemplate = QObject::tr("urn:xccdf:fix:script:sh");
```
mpreisler: use initialization lists, do this in the cpp file, not in the header.

src/RemediationRoleSaver.cpp
```diff
@@ -1,3 +1,24 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authors:
+ *      Martin Preisler <mpreisle@redhat.com>
```
mpreisler: ehm... nope :)

include/RemediationRoleSaver.h
```diff
@@ -19,7 +40,7 @@ extern "C"
 class RemediationSaverBase
 {
     public:
-	RemediationSaverBase(QWidget* parentWindow, ScanningSession* session): mParentWindow(parentWindow), mScanningSession(session) {}
+    RemediationSaverBase(QWidget* parentWindow, ScanningSession* session): mParentWindow(parentWindow), mScanningSession(session) {}
```
mpreisler: indents borked

src/RemediationRoleSaver.cpp
```diff
@@ -41,3 +62,21 @@ QString RemediationSaverBase::guessFilenameStem()
     // TODO: Add guess that uses benchmark and profile names
     return QString("remediation");
 }
+
+
+BashRemediationSaver::BashRemediationSaver(QWidget* parentWindow, ScanningSession* session):RemediationSaverBase(parentWindow, session)
+{
```
mpreisler: split the line above into multiple lines, break after `:` makes it more readable, indent the initialization list items and keep them one per line.

src/RemediationRoleSaver.cpp
```diff
@@ -41,3 +62,21 @@ QString RemediationSaverBase::guessFilenameStem()
...
+    mSaveMessage = QObject::tr("Save remediation role as a bash script");
+    mFiletypeExtension = "sh";
+    mFiletypeTemplate = QObject::tr("bash script (*.%1)");
+    mFixTemplate = QObject::tr("urn:xccdf:fix:script:sh");
```
mpreisler: any reason why this is not initialized in the initialization list?

matejak: It belongs to the base class and the base class doesn't know how should the content be.
Having it in the cpp file like this has the added benefit that If there is a new format of remediation rules, the person who will implement it will immediately know what is the meaning of each individual string thanks to the fact that it is assigned to the variable as opposed to being specified as "the 3rd argument to the constructor".

src/MainWindow.cpp
```diff
@@ -1531,3 +1551,16 @@ QMessageBox::StandardButton MainWindow::openNewFileQuestionDialog(const QString&
           QMessageBox::Yes | QMessageBox::No, QMessageBox::No
     );
 }
+
+// TODO: Reuse code between generateBashRemediationRole and generateAnsibleRemediationRole
```
mpreisler: outdated comment

src/RemediationRoleSaver.cpp
```diff
@@ -0,0 +1,82 @@
...
+        mFiletypeTemplate.arg(mFiletypeExtension), 0
+#ifndef SCAP_WORKBENCH_USE_NATIVE_FILE_DIALOGS
+        , QFileDialog::DontUseNativeDialog
+#endif
+    );
+
+    if (filename.isEmpty())
+        return;
+
+    int result = SaveToFile(filename);
```
mpreisler: can be const

src/RemediationRoleSaver.cpp
```diff
@@ -0,0 +1,82 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authors:
+ *      Martin Preisler <mpreisle@redhat.com>
+ */
+
+#include "RemediationRoleSaver.h"
+
+void RemediationSaverBase::SelectFilenameAndSaveRole()
```
mpreisler: method name

src/RemediationRoleSaver.cpp
```diff
@@ -0,0 +1,82 @@
...
+    {
+        // TODO: if OK
+    }
+    else
+    {
+        // TODO: if not OK
+    }
+}
+
+int RemediationSaverBase::SaveToFile(const QString& filename)
```
mpreisler: method name doesn't follow coding style

src/RemediationRoleSaver.cpp
```diff
@@ -0,0 +1,82 @@
...
+    QFile outputFile(filename);
+    outputFile.open(QIODevice::WriteOnly);
+    struct xccdf_session* session = mScanningSession->getXCCDFSession();
+    struct xccdf_policy* policy = xccdf_session_get_xccdf_policy(session);
+    int result = xccdf_policy_generate_fix(policy, NULL, mFixTemplate.toUtf8(), outputFile.handle());
+    outputFile.close();
+    return result;
+}
+
+QString RemediationSaverBase::guessFilenameStem()
```
mpreisler: this method can be const

src/RemediationRoleSaver.cpp
mpreisler: ah, ok, I didn't realize you are setting parent class members in the ctor

include/RemediationRoleSaver.h
```diff
@@ -0,0 +1,75 @@
...
+#include <xccdf_policy.h>
+#include <xccdf_session.h>
+}
+
+
+class RemediationSaverBase
+{
+    public:
+        RemediationSaverBase(QWidget* parentWindow, ScanningSession* session):
+            mParentWindow(parentWindow), mScanningSession(session) {}
```
mpreisler: move ctor to cpp

src/MainWindow.cpp
```diff
@@ -247,6 +249,24 @@ MainWindow::MainWindow(QWidget* parent):
...
+    QObject::connect(
+        genAnsibleRemediation, SIGNAL(triggered()),
+        this, SLOT(generateAnsibleRemediationRole())
+    );
+
+    QMenu* remediationButtonMenu = new QMenu(this);
+    remediationButtonMenu->addAction(genBashRemediation);
+    remediationButtonMenu->addAction(genAnsibleRemediation);
+    // remediationButtonMenu->addAction(mGenPuppetRemediation);
+    // remediationButtonMenu->addAction(mGenAnacondaRemediation);
```
mpreisler: the puppet command probably makes sense but anaconda roles don't have a use-case. I suggest to uncomment what's supposed to be there and remove anaconda action for good.

src/RemediationRoleSaver.cpp
```diff
@@ -0,0 +1,82 @@
...
+    }
+}
+
+int RemediationSaverBase::saveToFile(const QString& filename)
+{
+    QFile outputFile(filename);
+    outputFile.open(QIODevice::WriteOnly);
+    struct xccdf_session* session = mScanningSession->getXCCDFSession();
+    struct xccdf_policy* policy = xccdf_session_get_xccdf_policy(session);
+    int result = xccdf_policy_generate_fix(policy, NULL, mFixTemplate.toUtf8(), outputFile.handle());
```
mpreisler: since we are writing C++ we should handle C errors where we use C and only there IMO. Let's check the error here and raise an exception. Leaking error codes outside of methods makes it a pain to use the API.

mpreisler:
ACK

PR #144, opened 2017-10-25
Result-based generation of remediation roles

matejak:
This PR complements #143 since it allows users to generate remediation roles based on local scan results.

Aspects of the PR that may feel disturbing:

- Although I believe that the idea behind RemediationSaver classes is in its essence good, its expression in C++ doesn't feel like that. I would like to use mixin classes, but that would result in lots of base classes and associated boilerplate. Templates have much less boilerplate, but they have other shortcomings (some boilerplate + need to include cpp files). Anyway, the code now uses inheritance without templates.

- The remediation generation is now handled by the `oscap` tool wrapper in a same way as the scan itself. I get a harmless, but annoying error during the operation (`.../scap-workbench-oscap.sh: line 100: TARGET_RESULTS_XCCDF: unbound variable`). I assume that you will get it too. If there is a quick fix for it, then you may advise me on how to proceed, otherwise I see two possibilities here:

  - ~fix the wrapper script,~
  - ~write another wrapper script,~
  - don't use wrapper scripts (DONE), or
  - figure out how to use libscap to generate remediations (DONE).

- ~A concept of "local scanning session" has been introduced - it is a data structure that holds temp files, so they are not deleted right after the scan finishes, and they are ready for the remediation generator.~
 Update: The `ResultViewer` keeps track of the scanning data, so there is no real need of the local scanning session at this stage.

Known aspects still to be addressed:

- ~How to deal with result-based remediations of remote scanning (currently it is not supported).~
- Refactoring the code that handles the subprocess execution and monitoring, so it can be shared between executing openscap for scanning and for remediations (it works now, but there is some code duplication).
- Should library-based remediation generation be prioritized over `oscap`-based? The library code needed in order to generate remediations is non-trivial.


mpreisler:
Maybe I am missing something but couldn't you make the code way simpler by just taking the ARF the local or remote scanner gives you and using the OpenSCAP API to generate the remediation role?

matejak:
I don't know what is simpler. I have done some more research and I have discovered that some of the issues with the approach using the `oscap` as another process can be fixed quite easily. I went for the subprocess approach because offline remediation and scanning also uses it.
I am also working on an approach using `libopenscap` directly, but I have troubles getting it to work at the moment.

matejak:
Tests don't pass because the compiler is old (it can't deal with the `override` keyword). Otherwise, remediations done by the library and those done by the subprocess should work.

mpreisler:
Couple of things:
- why do we have both libopenscap and subprocess results remediation role implementations? don't they both do the same thing?
- please get rid of the override keyword, it is neat and all but it breaks the build on old compilers. as a general rule of thumb we never change requirements in a stable branch

matejak:
Those two branches do the same, and both do the job. I don't know which one should be the default, because each one has its drawbacks:

- The code using the library is difficult to read, because the library provides only the lowlevel layer. 
- The code using `oscap` executes a process, which has its issues, but the code expresses the intent behind and other operations performed by the workbench also use this approach.

Finally, I also don't know whether to keep one and completely discard the other one, or whether to keep both in the source code. If it was up to me, I would go for keeping both ways in the code.

matejak:
So the test server decided - since it has the `openscap 1.2.14` installed and the library-powered remediation generation uses `openscap 1.2.16` API, library-powered remediation generation has to be disabled by default, because it wouldn't pass the test.

src/RemediationRoleSaver.cpp
```diff
@@ -79,35 +131,192 @@ void RemediationSaverBase::saveToFile(const QString& filename)
...
+    TemporaryDir workingDir;
+    process.setWorkingDirectory(workingDir.getPath());
+    QString program("oscap");
+
+    process.start(program, args);
+    process.waitForStarted();
+
+    unsigned int pollInterval = 100;
+
+    while (!process.waitForFinished(pollInterval))
```
WesleyCeraso: You can get the same behavior by passing -1 as timeout.

src/RemediationRoleSaver.cpp
```diff
@@ -79,35 +131,192 @@ void RemediationSaverBase::saveToFile(const QString& filename)
...
+
+    // "" for profile should work, as it works when passed to the command-line oscap
+    if (xccdf_session_build_policy_from_testresult(session, "") != 0)
+        throw std::runtime_error("Couldn't get build policy from testresult");
+
+    struct xccdf_policy* policy = xccdf_session_get_xccdf_policy(session);
+    struct xccdf_result* result = xccdf_policy_get_result_by_id(policy, xccdf_session_get_result_id(session));
+    /* Result-oriented fixes */
+
+    QString role_template("urn:xccdf:fix:script:%1");
```
mpreisler: do this on one line, make it const

src/RemediationRoleSaver.cpp
```diff
@@ -79,35 +131,192 @@ void RemediationSaverBase::saveToFile(const QString& filename)
...
+    role_template = role_template.arg(mFixType);
+
+    QFile outputFile(filename);
+    outputFile.open(QIODevice::WriteOnly);
+    // Generate fix
+    int rc = xccdf_policy_generate_fix(policy, result, role_template.toUtf8().constData(), outputFile.handle());
```
mpreisler: const

include/RemediationRoleSaver.h
```diff
@@ -22,52 +22,143 @@
 #ifndef SCAP_WORKBENCH_REMEDIATION_ROLE_SAVER_H_
 #define SCAP_WORKBENCH_REMEDIATION_ROLE_SAVER_H_
 
-#include "ForwardDecls.h"
```
mpreisler: in Workbench we include ForwardDecls in every file as the first include on purpose. You never know what's going to be added to that file.

include/RemediationRoleSaver.h
```diff
@@ -22,52 +22,143 @@
...
-#include "ScanningSession.h"
-
 #include <QString>
 #include <QFileDialog>
 
+#include "ScanningSession.h"
+#include "OscapScannerLocal.h"
```
mpreisler: is there any reason why you changed the order of includes?

src/OscapScannerLocal.cpp
```diff
@@ -117,34 +87,38 @@ void OscapScannerLocal::evaluate()
     }
 
     // TODO: Error handling!
-    emit infoMessage(QObject::tr("Creating temporary files..."));
+    // TODO: vvv Move to a relevant place vvv
+    // emit infoMessage(QObject::tr("Creating temporary files..."));
```
mpreisler: TODO in the code

src/OscapScannerLocal.cpp
```diff
@@ -166,17 +140,7 @@ void OscapScannerLocal::evaluate()
...
-#ifdef SCAP_WORKBENCH_LOCAL_NICE_FOUND
-    args.prepend(getPkexecOscapPath());
-    args.prepend(QString::number(SCAP_WORKBENCH_LOCAL_OSCAP_NICENESS));
-    args.prepend("-n");
-
-    program = SCAP_WORKBENCH_LOCAL_NICE_PATH;
-#else
-    program = getPkexecOscapPath();
-#endif
+    QString program = getOscapProgram(args);
```
mpreisler: const, the name of the method could be better but that's not a big deal

src/RemediationRoleSaver.cpp
```diff
@@ -30,12 +30,36 @@ extern "C"
 #include <xccdf_benchmark.h>
 #include <xccdf_policy.h>
 #include <xccdf_session.h>
+#include <ds_rds_session.h>
```
mpreisler: AFAIK ds_rds_session is only exposed in recent OpenSCAP, we should find the exact version and have a macro guard around this

src/RemediationRoleSaver.cpp
```diff
@@ -79,35 +131,192 @@ void RemediationSaverBase::saveToFile(const QString& filename)
...
+
+    args.append(mArfFile.fileName());
+
+    // TODO: Launching a process and going through its output is something we do already
+    // This is a lightweight launch though.
+    QProcess process(RemediationSaverBase::mParentWindow);
+
+    TemporaryDir workingDir;
+    process.setWorkingDirectory(workingDir.getPath());
+    QString program("oscap");
```
mpreisler: NACK, this should use the configure time OSCAP path, we can't hardcode any paths, this would never work on Windows or MacOSX

src/RemediationRoleSaver.cpp
```diff
@@ -79,35 +131,192 @@ void RemediationSaverBase::saveToFile(const QString& filename)
...
+
+    process.start(program, args);
+    process.waitForStarted();
+
+    unsigned int pollInterval = 100;
+
+    while (!process.waitForFinished(pollInterval))
+    {}
```
mpreisler: shouldn't you be calling the event pump in the loop?

src/RemediationRoleSaver.cpp
```diff
@@ -79,35 +131,192 @@ void RemediationSaverBase::saveToFile(const QString& filename)
...
+    if (session == NULL)
+        throw std::runtime_error("Couldn't get XCCDF session from the report source");
+
+    xccdf_session_set_loading_flags(session, XCCDF_SESSION_LOAD_XCCDF);
+    if (xccdf_session_load(session) != 0)
+        throw std::runtime_error("Couldn't get load XCCDF");
+
+    xccdf_session_set_loading_flags(session, XCCDF_SESSION_LOAD_XCCDF);
+    if (xccdf_session_load(session) != 0)
+        throw std::runtime_error("Couldn't load XCCDF");
```
mpreisler: why are you doing this twice?

src/RemediationRoleSaver.cpp
```diff
@@ -79,35 +131,192 @@ void RemediationSaverBase::saveToFile(const QString& filename)
...
-    mFixTemplate = QObject::tr("urn:xccdf:fix:script:puppet");
+    struct oscap_source* source = oscap_source_new_from_file(mArfFile.fileName().toUtf8().constData());
+    oscap_document_type_t document_type = oscap_source_get_scap_type(source);
+    if (document_type != OSCAP_DOCUMENT_ARF) {
+        throw std::runtime_error("Expected an ARF file");
+    }
+
+    struct xccdf_session* session;
+    struct ds_rds_session* arf_session;
+    arf_session = ds_rds_session_new_from_source(source);
```
mpreisler: declare and assign on one line

src/RemediationRoleSaver.cpp
```diff
@@ -79,35 +131,192 @@ void RemediationSaverBase::saveToFile(const QString& filename)
...
+    struct oscap_source *report_source = ds_rds_session_select_report(arf_session, NULL);
+    if (report_source == NULL) {
+        throw std::runtime_error("Couldn't get report source from the ARF session");
+    }
+    struct oscap_source *report_request_source = ds_rds_session_select_report_request(arf_session, NULL);
+    if (report_request_source == NULL) {
+        throw std::runtime_error("Couldn't get report request source from the ARF session");
+    }
+
+    session = xccdf_session_new_from_source(oscap_source_clone(report_request_source));
```
mpreisler: declare and assign the session here.

mpreisler:
I have done a full review of the codebase, besides the one blocker - you need to use the configure time oscap path and avoid hard-coding - I would be OK with merging it as is. Added a bunch more comments to make the code base better but these are not strictly required.

src/RemediationRoleSaver.cpp
matejak: The functionality for this is tightly coupled to the `OscapScannerLocal` and given the nature of this process, I think that it is not worth the effort - checking for hangs and return codes should be enough as errors are extremely unlikely.

matejak:
I think that I have addressed all reviewer's comments (except polling of the `oscap` process when generating result-based remediations).

src/RemediationRoleSaver.cpp
```diff
@@ -79,35 +135,192 @@ void RemediationSaverBase::saveToFile(const QString& filename)
...
-    mSaveMessage = QObject::tr("Save remediation role as a bash script");
-    mFiletypeExtension = "sh";
-    mFiletypeTemplate = QObject::tr("bash script (*.%1)");
-    mFixTemplate = QObject::tr("urn:xccdf:fix:script:sh");
+    QStringList args;
+    args.append("xccdf");
+    args.append("generate");
+    args.append("fix");
+
+    args.append("--fix-type");
```
mpreisler: NACK, this is a big one, use --template, not fix type! --template works with old versions of OpenSCAP, fix-type is very recent.

mpreisler:
Found one more blocker, you are using --fix-type which is only available in openscap 1.2.16, yet you are not checking oscap version. If you used --template it would work with all reasonably recent versions of openscap.

src/ResultViewer.cpp
```diff
@@ -55,6 +58,28 @@ ResultViewer::ResultViewer(QWidget* parent):
     mSaveMenu->addAction(mSaveReportAction);
     mUI.saveButton->setMenu(mSaveMenu);
 
+    // TODO: Are the menu and actions neccessary as class atributes?
```
mpreisler: seems like an outdated comment

src/ResultViewer.cpp
```diff
@@ -55,6 +58,28 @@ ResultViewer::ResultViewer(QWidget* parent):
...
+    QAction* mGenBashRemediation = new QAction("&bash", this);
```
mpreisler: local variables can't start with m. m is for members

src/ResultViewer.cpp
```diff
@@ -98,6 +123,11 @@ void ResultViewer::loadContent(Scanner* scanner)
 
     mARF.clear();
     scanner->getARF(mARF);
+
+    mCurrentLocalScanner = NULL;
```
mpreisler: already set in initializer list

src/RemediationRoleSaver.cpp
```diff
@@ -79,35 +133,190 @@ void RemediationSaverBase::saveToFile(const QString& filename)
...
+    // vvv This will work, if there is only one result ID in the ARF file, it will be picked no matter what the argument value is.
+    // However, ommitting --result-id "" won't work.
+    args.append("--result-id");
+    args.append("");
+
+    args.append(mArfFile.fileName());
+
+    // Launching a process and going through its output is something we do already in OscapScannerLocal::evaluate()
+    // This is a lightweight launch though.
+    QProcess process(RemediationSaverBase::mParentWindow);
```
mpreisler: you don't need the `RemediationSaverBase::` prefix here.

src/RemediationRoleSaver.cpp
```diff
@@ -79,35 +133,190 @@ void RemediationSaverBase::saveToFile(const QString& filename)
...
+
+    TemporaryDir workingDir;
+    process.setWorkingDirectory(workingDir.getPath());
+    QString program(SCAP_WORKBENCH_LOCAL_OSCAP_PATH);
+
+    process.start(program, args);
+    process.waitForStarted();
+
+    const unsigned int remediation_generation_timeout = 10000;
```
mpreisler: code style :-/

mpreisler:
Some of the things I commented before creeped back in. the member variable naming scheme shouldn't be used for local variables, that is super confusing. Plus commented a bunch of minor things.

Once the member vars are fixed I am OK with merging this, feel free to ask somebody else to merge if I am not available, I am not in a good timezone for collab right now :(

We should definitely update user manual before release, are we tracking that anywhere? There are a lot of GUI changes.

mpreisler:
Tried it on my Fedora, empty profile with one passing and one failing rule. When I click generate remediation role->bash I get nothing in the GUI but on stderr this appears:
```
18:12:18 | info     | Processing has been finished!
Error saving remediation role: There was an error in course of remediation role generation! Exit code of the 'oscap' process was 1.
```

mpreisler:
One more thing, I noticed that both from profile and from results role generation is called "Generate remediation role". This may be confusing. Not everyone will hover for a tooltip to figure out the difference. I also noticed that the from profile button is way smaller than the from results. Lastly, the button changes position, it's all the way on the left when generating from profile, it's second in the right group when generating from results.

mpreisler:
image

Nothing in diagnostics, no error message or dialog shown, only the stderr output in the terminal. Keep in mind that almost nobody runs GUI tools with stdout/stderr attached, so these messages get dropped in 99% of all cases.

mpreisler:
image

vs 

image

I think we should be more consistent in the UX.

mpreisler:
Just noticed that the from profile generation has the same issue, Try saving it in /root or /boot where you don't have permission. SCAP Workbench will not show any errors, it all looks like it worked. It will only output something to stderr.

matejak:
Concerning deletion of files:
I was somehow reluctant to do that. I agree that empty files that happen to be there after `oscap` finishes should be deleted, but I consider the root cause to lie within the `oscap` process that creates those empty files.

Concerning the button position: I gave it some thought when designing the layout. I considered the lower right corner to be "hot" - it should contain buttons that the user is likely to use.
* In the main window, user is very likely to scan and it is very unlikely that he would be interested in remediation roles that contain fixes for problems he doesn't have, therefore the button is far to the left.
* Conversely, when finishing a scan, one surely appreciates to like to see the scanning results in a concise from. Fixing them is probably the second most important thing. Scan results are not human-readable and therefore less attractive for the user, so I placed the button next to the "show report" one.

Concerning the button label: I was thinking of "save profile/result-based fixes" ("fixes " are much shorter than "remediation roles"), but it would introduce inconsistency in naming. A label saying "Generate profile-based remediation roles" is way too long and I can't think how to abbreviate it in a smart way.

src/RemediationRoleSaver.cpp
```diff
@@ -301,6 +318,10 @@ void ResultBasedLibraryRemediationSaver::saveToFile(const QString& filename)
     outputFile.open(QIODevice::WriteOnly);
 
     const int rc = xccdf_policy_generate_fix(policy, result, mTemplateString.toUtf8().constData(), outputFile.handle());
+    if (!outputFile.flush())
+    {
+        throw std::runtime_error("Could not write to the destination location.");
```
mpreisler: This is not what I had in mind. I was just giving an example when it fails, patching just the symptom I pointed out instead of the root cause is suboptimal. IMO we should report oscap errors well via a dialog + stderr into workbench diagnostics dialog. That would solve all of these issues.

You absolutely don't want to flush the file from Qt btw, you never wrote anything to the file via the Qt interface and you opened it write only. Whether Qt empties the file or not is implementation defined.

src/RemediationRoleSaver.cpp
```diff
@@ -92,6 +92,18 @@ void RemediationSaverBase::selectFilenameAndSaveRole()
...
+void RemediationSaverBase::removeFileWhenEmpty(const QString& filename)
+{
+    QFile outputFile(filename);
+    outputFile.open(QIODevice::ReadOnly);
+    int fileSize = outputFile.size();
+    outputFile.close();
+
+    if (fileSize == 0)
+        QFile::remove(filename);
+}
```
mpreisler: IMO we can avoid all of this, this is added complexity that the rest of the workbench doesn't do anyway.

src/RemediationRoleSaver.cpp
matejak: The problem this addresses are cases when the workbench wants to write zero-sized files to directories that are not writable. Currently, there are no `puppet` remediations, so when I want to save a remediation profile, there is no error, but the result is an empty file. Then:
* If the destination is writable, an empty file is created and stays there, whereas
* if the destination is not writable, there is no error even if no file is created. Errors happen during the `flush`, calling `outputFile.close()` silences any `flush` errors.
This inconsistency that I have discovered thank to your suggestion seemed misleading to me in to an extent that it justified this ad-hoc fix. Anyway, should I keep this, or should I remove this?

src/RemediationRoleSaver.cpp
matejak: I got the impression that you would like to delete the empty files after a failed generation attempt.
Moreover - I think that creating empty files is unavoidable when using the library. I have to open the file, pass the descriptor to the function that does the generation, then I have to close the file regardless whether the generation went well or not - this produces either a remediation role, or an empty file.

matejak:
I have difficulties with error messages - the code is structured in such way that it is easy log errors when one has link to the main window. From what I was able to see, accessing the diagnostics dialog from the results viewer would require either substantial refactoring or introduction of ugly hacks. Therefore, I decided to go for something simpler s.a. displaying a [GUI error message](https://github.com/OpenSCAP/scap-workbench/pull/144/files#diff-955a7e39408415a7b86ee7cb672d89f1R126), but it doesn't show up.

mpreisler:
@matejak Throw exceptions and catch them in MainWindow?

src/RemediationRoleSaver.cpp
```diff
@@ -53,22 +83,76 @@ void RemediationSaverBase::selectFilenameAndSaveRole()
...
+    if (mDiagnostics != NULL)
+    {
+        mDiagnostics->errorMessage(completeErrorMessage);
+    }
+    else
+    {
+        std::cerr << completeErrorMessage.toUtf8().constData();
+
+        // TODO: This doesn't work for some reason, even if mParentWindow points to ResultViewer instance
+        QErrorMessage qErrorMsg(mParentWindow);
```
mpreisler: because it goes out of scope and is deallocated... but this is a no-go regardless, let's not invent more ways to show errors in SCAP Workbench, let's instead use what we already have

matejak:
Reporting errors from remediation generation code in a unified way turned out to be very tricky. After exploring several possibilities together with @WesleyCeraso , we came to a conclusion that there is no clear winner among them.
I have decided to introduced a global object `globalDiagnosticsDialog`  for these purposes. It effectively is a singleton, `mDiagnosticsDialog` in `MainWindow` is the same thing. Since the class is [an interface to a write-only stateless logging service](https://stackoverflow.com/a/6717265), I don't consider introducing this global object harmful.

src/RemediationRoleSaver.cpp
```diff
@@ -79,35 +151,194 @@ void RemediationSaverBase::saveToFile(const QString& filename)
...
+
+    args.append("--template");
+    args.append(mTemplateString);
+    args.append("--output");
+    args.append(filename);
+
+    // vvv This will work, if there is only one result ID in the ARF file, it will be picked no matter what the argument value is.
+    // However, ommitting --result-id "" won't work.
+    args.append("--result-id");
+    args.append("");
```
jan-cerny: Result-id should not be "". I wonder why it works, it looks like a hack.

src/RemediationRoleSaver.cpp
```diff
@@ -79,35 +151,194 @@ void RemediationSaverBase::saveToFile(const QString& filename)
...
+    if (!process_finished_on_time)
+    {
+        QString message = QObject::tr("The process that was supposed to generate remediations didn't finish on time (i.e. within %1 secs), so it was terminated.").arg(remediationGenerationTimeout / 1000);
+        process.kill();
+        throw std::runtime_error(message.toUtf8().constData());
+    }
+
+    if (process.exitCode() != 0)
+    {
+        throw std::runtime_error(QObject::tr("There was an error in course of remediation role generation! Exit code of the 'oscap' process was %1.").arg(process.exitCode()).toUtf8().constData());
```
jan-cerny: Maybe it could be nice to copy the oscap's stderr to make the report more specific.

jan-cerny:
When I want to generate a remediation from a tailored profile, it doesn't work. I think it needs to add `--tailoring-file` to `oscap xccdf generate fix` command.

I have a filed an issue to OpenSCAP, because ARF doesn't contain the tailoring. See https://github.com/OpenSCAP/openscap/issues/902

src/RemediationRoleSaver.cpp
matejak: This is difficult to tell. The `--result-id` is in fact a result suffix, And an empty string is a suffix of every result ID, which is convenient as it is guaranteed that we have only one result per ARF.
I think that having this as part of the official `oscap` API is better than implementing the tedious result ID extraction here in the workbench.

src/RemediationRoleSaver.cpp
matejak: I agree, but there is the code that does that already in the workbench.
Unfortunately, it is interwoven in the `OscapScannerLocal` instance and used in the `OscapScannerLocal::evaluate` method in a way that it is not easy to extract and use elsewhere.
I propose to leave it like it is for now and make the clean fix as another PR.

jan-cerny: @matejak OK, I agree, let's fix that later in another pull request. Thank you for investigation.

matejak: I have squeezed a simple fix good-enough for the time being.

matejak:
Good catch, but this will be a bit difficult. Proper fix requires access to the scanning session, which is the property of `MainWindow`. The remediation generation is triggered from the `ResultsViewer` window.
Fixing this would require decoupling UI from the scanning according to the MVC pattern, which is out of scope for this PR. I view this issue as a non-blocker, although there should be a workbench issue that keeps track of this too.

jan-cerny:
@matejak OK, doing that in a separate pull request is a great idea. Could you file an issue on SCAP Workbench describing that problem?

src/DiagnosticsDialog.cpp
```diff
@@ -164,3 +164,5 @@ void DiagnosticsDialog::copyToClipboard()
     QClipboard* clipboard = QApplication::clipboard();
     clipboard->setText(fullLog);
 }
+
+DiagnosticsDialog* globalDiagnosticsDialog = NULL;
```
mpreisler: make this a proper documented singleton, static member variables inside DiagnosticsDialog, remove the variable from MainWindow, make everything use it via the singleton var.

src/RemediationRoleSaver.cpp
```diff
@@ -79,35 +151,195 @@ void RemediationSaverBase::saveToFile(const QString& filename)
...
+    outputFile.close();
+    ds_rds_session_free(arf_session);
+    xccdf_session_free(session);
+
+    if (rc != 0)
+    {
+        const char* err = oscap_err_desc();
+        if (err == NULL)
+            err = "Unknown error";
+        throw std::runtime_error(err);
```
mpreisler: what happens if this throws an exception? It seems to me that the whole app would crash.

src/RemediationRoleSaver.cpp
```diff
@@ -79,35 +151,195 @@ void RemediationSaverBase::saveToFile(const QString& filename)
...
+
+    struct xccdf_policy* policy = xccdf_session_get_xccdf_policy(session);
+    struct xccdf_result* result = xccdf_policy_get_result_by_id(policy, xccdf_session_get_result_id(session));
+    /* Result-oriented fixes */
+
+    QFile outputFile(filename);
+    outputFile.open(QIODevice::WriteOnly);
+
+    const int rc = xccdf_policy_generate_fix(policy, result, mTemplateString.toUtf8().constData(), outputFile.handle());
+    if (!outputFile.flush())
```
mpreisler: doesn't this flush the Qt cache? Is this defined in the docs? We are using the handle directly and AFAIK Qt guarantees that the file will exist when we call "handle()", however I don't think flush() behavior is well defined for this case.

matejak: [The docs}(http://doc.qt.io/qt-4.8/qfile.html#close) tell that `close()` calls `flush()`, but ignores any errors. I have verified that if there is nothing to write, the `xccdf_policy_generate_fix` doesn't complain, and `close()` doesn't indicate anything either. The only way how to verify that everything went OK seems to be to call `flush` manually, check for errors and then proceed with `close()`.

src/RemediationRoleSaver.cpp
matejak: No, it is caught up in `selectFilenameAndSaveRole` like all other exceptions. I have verified it and it works.

src/DiagnosticsDialog.cpp
matejak: I have no objection to document it, but I think that the rest is out of scope for this PR. The need for the singleton comes from tight coupling between the scanning logic and UI. As this coupling causes another problems, I would rather spend some time by decoupling that than making this fix nicer.
I see this as a matter of discussion, but even if we decide to use the singleton and make the whole code use it, it would fit to another PR perfectly.

matejak: I have tried the signal approach before, but I have failed - I am a QT signal newbie and there was nobody to ask at that time: Objects can send signals to the dialog thanks to the fact that they are connected to it (see src/MainWindow.cpp:630). MainWindow can connect anything, but there is no way how to present instances of remediation savers to it, ResultsViewer could connect something, but it doesn't know of the diagnostics dialog.
I have not explored the signal way further, because there seemed to be a better way.

mpreisler: I can help you with this, the idea is to signal errorMessage with a string and connect that. Passing the saver ref shouldn't be necessary.

mpreisler: Check out https://github.com/OpenSCAP/scap-workbench/blob/v1-1/src/MainWindow.cpp#L634

include/OscapScannerBase.h
```diff
@@ -44,6 +44,10 @@ class OscapScannerBase : public Scanner
         virtual void getReport(QByteArray& destination);
         virtual void getARF(QByteArray& destination);
 
+        static QStringList createRemediationRole(const QString& arfFile,
+                                                 const QString& fix_type,
```
mpreisler: code style

src/DiagnosticsDialog.cpp
matejak: We have agreed that this way of solving the problem by passing signals is indeed not an improvement, so properly documenting the global variable will be enough for the time being.

src/RemediationRoleSaver.cpp
```diff
@@ -24,24 +24,53 @@
...
+
+const QString puppetSaveMessage = QObject::tr("Save remediation role as a puppet manifest");
+const QString puppetFiletypeExtension = "pp";
+const QString puppetFiletypeTemplate = QObject::tr("puppet manifest (*.%1)");
+const QString puppetFixType = QString("puppet");
+
+
+RemediationSaverBase::RemediationSaverBase(QWidget* parentWindow,
+                const QString& saveMessage, const QString& filetypeExtension, const QString& filetypeTemplate, const QString& fixType):
+    mParentWindow(parentWindow), mDiagnostics(globalDiagnosticsDialog), mSaveMessage(saveMessage), mFiletypeExtension(filetypeExtension), mFiletypeTemplate(filetypeTemplate),
```
mpreisler: why do you create another member and set it to globalDiagnosticsDialog?

src/RemediationRoleSaver.cpp
mpreisler: I think this is a little excessive, I was just using this example to demonstrate that there is no error reporting. It can probably stay though..

src/RemediationRoleSaver.cpp
```diff
@@ -53,22 +82,65 @@ void RemediationSaverBase::selectFilenameAndSaveRole()
...
+}
+
+
+void RemediationSaverBase::saveFileOK(const QString& filename)
+{
+    // TODO: if OK - inform the user
+}
+
+
+void RemediationSaverBase::saveFileError(const QString& filename, const QString& errorMsg)
```
mpreisler: why are there 2 newlines between methods everywhere? shouldn't it be 2 lines between different classes but 1 line between methods?

src/RemediationRoleSaver.cpp
matejak: A part of the class is isolated from the fact that there is a global variable out there. Getting rid of it will be slightly easier thanks to this separation.

mpreisler:
ACK. Great stuff!

PR #149, opened 2017-11-14
pkg-config path needed if openscap is compiled locally

openprivacy:
As mentioned recently on the mailing list:
> I've compiled openscap and the SSG locally and they work great. I'd like to try scap-workbench but cmake complains "No package 'libopenscap' found" despite having run "make install" on openscap.

Openscap comes with a `libopenscap.pc` file, but `pkg-config` can't find it. The fix is simple and included here in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,9 @@ optional dependencies:
 2) Build SCAP Workbench:
 ```console
 $ mkdir build; cd build
+$ # this export line is only required if openscap was
+$ # compiled locally and libopenscap is in /usr/local/lib/
+$ export PKG_CONFIG_PATH=/PATH/TO/YOUR/SOURCE/openscap/
 $ cmake ../
 $ make
 ```
--- a/README.md
+++ b/README.md
@@ -36,8 +36,8 @@ optional dependencies:
 ```console
 $ mkdir build; cd build
 $ # This export line is only required if openscap was compiled locally;
-$ # add to PKG_CONFIG_PATH the directory containing libopenscap.pc
-$ export PKG_CONFIG_PATH=$PKG_CONFIG_PATH:/PATH/TO/DIR/WITH/.pcFILE/
+$ # add to PKG_CONFIG_PATH the directory containing installed libopenscap.pc
+$ export PKG_CONFIG_PATH="$PKG_CONFIG_PATH:/PATH/TO/DIR/WITH/.pcFILE/"
 $ cmake ../
 $ make
 ```
--- a/README.md
+++ b/README.md
@@ -35,9 +35,9 @@ optional dependencies:
 2) Build SCAP Workbench:
 ```console
 $ mkdir build; cd build
-$ # this export line is only required if openscap was
-$ # compiled locally and libopenscap is in /usr/local/lib/
-$ export PKG_CONFIG_PATH=/PATH/TO/YOUR/SOURCE/openscap/
+$ # This export line is only required if openscap was compiled locally;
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -35,6 +35,9 @@ optional dependencies:
 2) Build SCAP Workbench:
 ```console
 $ mkdir build; cd build
+$ # this export line is only required if openscap was
+$ # compiled locally and libopenscap is in /usr/local/lib/
+$ export PKG_CONFIG_PATH=/PATH/TO/YOUR/SOURCE/openscap/
```
matejak: I think that the `PKG_CONFIG_PATH` should point the the location where the installed `.pc` files will be, which is by default `<prefix>/lib/pkgconfig`.
For your example, that would yield `/usr/local/lib/pkgconfig`, for standard user installs `$HOME/.local/lib/pkgconfig`. Do you agree? I am not saying that your take can't work, but `pkg-config` would pick up the `.pc` file generated during the build and not yet installed. Picking the installed one is much better.

openprivacy: Your suggestion would be better, but perhaps I then need to do a PR to openscap, as `make install` does not install the .pc file (you are right that it should be installed in `/usr/local/lib/pkgconfig/`).

mpreisler: Agreed. the PKG_CONFIG_PATH solution is not the correct one and it will cause the configuration to find openscap but miss other packages. I think there are better solutions than altering PKG_CONFIG_PATH but in case we decide to go with this it should append to the path instead of replacing it.

ex: `export PKG_CONFIG_PATH=/PATH/TO/YOUR/SOURCE/openscap/:$PKG_CONFIG_PATH`

matejak: I have two remarks:
* In my case, `make install` installs the `libopenscap.pc` file to `<prefix>/lib/pkgconfig`, at least if if the `prefix` is `$HOME/.local`. I can confirm it by removing `$HOME/.local/lib/pkgconfig/libopenscap.pc` and making it reappear after running `make install` in the openscap build directory.
* `PKG_CONFIG_PATH` is the correct variable to set. According to the `pkg-config` manual, it is the secondary `.pc` path. This means that if one overrides it during build of other packages, primary `pkg-config` search directories are still searched through. Martin's proposal is the best - it produces an expected result even if the user has `PKG_CONFIG_PATH` already set e.g. via a modified `.bashrc`.

openprivacy:
@mpreisler agreed; updated.
@matejak Maybe it requires `prefix` to be set, as when I `sudo make install` code gets installed in `/usr/local` but the `/usr/local/lib/pkgconfig` dir is not created.

I made some changes that should help?

README.md
```diff
@@ -35,6 +35,9 @@ optional dependencies:
 2) Build SCAP Workbench:
 ```console
 $ mkdir build; cd build
+$ # This export line is only required if openscap was compiled locally;
+$ # add to PKG_CONFIG_PATH the directory containing libopenscap.pc
+$ export PKG_CONFIG_PATH=$PKG_CONFIG_PATH:/PATH/TO/DIR/WITH/.pcFILE/
```
matejak: Please consider these adjustments:
* subsitute "containing libopenscap.pc" with something like "containing _installed_ libopenscap.pc _(typically located in a directory named 'pkgconfig')_", so users can search for it and find the right one quickly, and
* double quote the assignment (like `X="$X:some other stuff"`) so it works even for directories containing spaces.

I have no further ideas how to improve your contribution before merging it.

matejak:
Great, thank you for your contribution!

PR #164, opened 2018-06-04
Reflect qt5-devel dependency

cipherboy:
Master branch requires `qt5-devel` versus `qt-devel` package to be installed; this reflects that.


mpreisler:
@openscap-ci add to whitelist

PR #166, opened 2018-06-06
Remove stray line in tailoring window

cipherboy:
It is a stray line in the tailoring window. (@mpreisler says that it was probably left over from when the delete button was moved left).


mpreisler:
ACK when jenkins finishes

mpreisler:
Actually, this is master so jenkins will fail here, it does not have qt5 installed yet.

PR #167, opened 2018-06-06
Add openssh-askpass as a required dependency

cipherboy:
Required for remote ssh scanning.


mpreisler:
Could you target v1-1, please? This is a bug there as well.

cipherboy:
Okay.

PR #170, opened 2018-06-06
Add openssh-askpass as a required dependency

cipherboy:
Required for remote ssh scanning.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ How to run it out of the box
 
 required dependencies:
 ```console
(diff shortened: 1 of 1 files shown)


cipherboy:
@mpreisler This one is against v1-1.

mpreisler:
ACK when jenkins finishes

mpreisler:
I canceled this jenkins job to save some resources. There is no way this can affect automated testing since it's a README change...

PR #173, opened 2018-06-07
-Werror -Wall -Wextra

cipherboy:
A perfectly clean build! Absolutely no warnings or errors during build.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.6)
+cmake_minimum_required(VERSION 2.8.6)
 
 project("scap-workbench")
 set(SCAP_WORKBENCH_VERSION_MAJOR "1")
@@ -9,6 +9,7 @@ set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake/")
 
 include(GNUInstallDirs)
 
+set(CMAKE_AUTOMOC ON)
 find_package(Qt5Widgets REQUIRED)
 find_package(Qt5XmlPatterns REQUIRED)
 
@@ -101,7 +102,6 @@ set(SCAP_WORKBENCH_SCAP_CONTENT_DIRECTORY "/usr/share/xml/scap" CACHE PATH
     "Directory where SCAP content is *usually* stored. This is the default directory when opening a file via file open dialog.")
 
 file(GLOB scap_workbench_HEADERS "${CMAKE_SOURCE_DIR}/include/*.h")
-qt5_wrap_cpp(scap_workbench_HEADERS_MOC ${scap_workbench_HEADERS})
 
 file(GLOB scap_workbench_UIS "${CMAKE_SOURCE_DIR}/ui/*.ui")
 qt5_wrap_ui(scap_workbench_UIS_HEADERS ${scap_workbench_UIS})
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -127,6 +127,15 @@ if (WIN32)
     configure_file("scap-workbench.wxs.in" "${CMAKE_CURRENT_BINARY_DIR}/scap-workbench.wxs")
 endif()
 
+# Enforce std=c++11 for old gcc versions to fix build in RHEL 7
+if (${CMAKE_CXX_COMPILER_ID} STREQUAL "GNU" AND ${CMAKE_CXX_COMPILER_VERSION} VERSION_LESS 4.9)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
+endif()
+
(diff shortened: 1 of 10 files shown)


src/TailoringWindow.cpp
```diff
@@ -131,11 +131,11 @@ TailoringWindow::TailoringWindow(struct xccdf_policy* policy, struct xccdf_bench
 
     QObject::connect(
         mUI.itemsTree, SIGNAL(currentItemChanged(QTreeWidgetItem*,QTreeWidgetItem*)),
-        this, SLOT(itemSelectionChanged(QTreeWidgetItem*,QTreeWidgetItem*))
+        this, SLOT(itemSelectionChanged(QTreeWidgetItem*))
     );
     QObject::connect(
-        mUI.itemsTree, SIGNAL(itemChanged(QTreeWidgetItem*,int)),
-        this, SLOT(itemChanged(QTreeWidgetItem*,int))
+        mUI.itemsTree, SIGNAL(itemChanged(QTreeWidgetItem*, int)),
```
cipherboy: Sorry about the spare whitespace change here. Could someone more familiar with the QT framework explain why it was necessary that this have the "int" parameter while it is marked unused later (on line 138)?

src/OscapScannerRemoteSsh.cpp
```diff
@@ -415,8 +415,11 @@ QString OscapScannerRemoteSsh::createRemoteTemporaryFile(bool cancelOnFailure)
                         "data to! Diagnostic info: %1").arg(proc.getDiagnosticInfo())
         );
 
-        mCancelRequested = true;
-        signalCompletion(mCancelRequested);
+        if (cancelOnFailure) {
```
cipherboy: No callers set `cancelOnFailure=false` (so the change makes no difference), but I believe this is the correct way to implement this behavior. I'd appreciate comments if its not.

src/RemediationRoleSaver.cpp
```diff
@@ -102,7 +103,10 @@ void RemediationSaverBase::removeFileWhenEmpty(const QString& filename)
 
 void RemediationSaverBase::saveFileOK(const QString& filename)
 {
-    // TODO: if OK - inform the user
+    QMessageBox::information(
```
cipherboy: `filename` was previously unused, so I added a information box. It helps with some of the "silent success" saves.

src/TailoringWindow.cpp
```diff
@@ -333,8 +333,8 @@ void TailoringWindow::synchronizeTreeItemSelections(QTreeWidgetItem* treeItem)
                 }
                 break;
             }
-            //fall through
         }
+            //fallthrough
```
cipherboy: This was necessary to appease the `gcc` gods. Any alternative suggestions would be appreciated.

src/OscapScannerBase.cpp
```diff
@@ -383,6 +383,9 @@ bool OscapScannerBase::tryToReadStdOutChar(QProcess& process)
                         emit warningMessage(QString("Failed to download \"%1\"!").arg(mLastDownloadingFile));
                 }
                 break;
+            default:
```
cipherboy: This has symmetry with the above (hidden in diff) block, and based upon it not affecting current operations, I'm going to assume this is correct, but I'd appreciate someone more familiar with the code to validate that it is.

CMakeLists.txt
```diff
@@ -127,6 +127,11 @@ if (WIN32)
     configure_file("scap-workbench.wxs.in" "${CMAKE_CURRENT_BINARY_DIR}/scap-workbench.wxs")
 endif()
 
+if (${CMAKE_C_COMPILER_ID} STREQUAL "GNU")
```
mpreisler: let's check CXX COMPILER and check it for GNU or Clang, Clang has the same options AFAIK.

mpreisler:
@openscap-ci test this please

mpreisler:
@openscap-ci test this please

cipherboy:
Wat?

```
/usr/include/c++/4.8.2/bits/c++0x_warning.h:32:2: error: #error This file requires compiler and library support for the ISO C++ 2011 standard. This support is currently experimental, and must be enabled with the -std=c++11 or -std=gnu++11 compiler options.
```

mpreisler:
> @mpreisler What type of system is this?

RHEL7

cipherboy:
@mpreisler Updated. Clang found another error (unused mSkipValid).

mpreisler:
@openscap-ci test this please

cipherboy:
@mpreisler AUTOMOC support: https://blogs.kde.org/2011/11/01/cool-new-stuff-cmake-286-automoc

mpreisler:
we need to raise cmake minimum required version to 2.8.6 because of automoc, it's the first line in CMakeLists.txt

src/TailoringWindow.cpp
mpreisler: I think we may need to keep the int. It's necessary for the slot signal connection to happen. moc needs it to choose the correct method. In essence you need to use full method names and since C++ supports overloading you need to include parameters.

mpreisler:
Finally... jenkins passed on RHEL7!

Issue #168, opened 2018-06-06
User/Host not saved when running remote scan


cipherboy:
When trying to perform a remote scan of a machine, the "User and host" input form is cleared every time "Scan" is pressed. The last-used value should be saved, e.g., for when a typo is introduced and the scan cannot be performed (`loclahost` instead of `localhost`) for better UX.

This seems to be an artifact of merging `v1-1` -> `master`; `git bisect` found this commit at fault: https://github.com/OpenSCAP/scap-workbench/commit/2266c7d29cd4ae716030f22cb1cf1bc7fb17044a in the merged tree.

cipherboy:
Closing as unable to reproduce any longer. See #175 for extended discussion.

PR #175, opened 2018-06-07
Reload ssh host more frequently

cipherboy:
This persists it and reloads in more places. This seems like the better place to put this code as it gets triggered on nearly all UI actions. I'm not entirely clear why it is getting cleared frequently otherwise. 

Closes: #168

### DIFF
--- a/src/RemoteMachineComboBox.cpp
+++ b/src/RemoteMachineComboBox.cpp
@@ -90,9 +90,6 @@ void RemoteMachineComboBox::notifyTargetUsed(const QString& target)
 
     syncToQSettings();
     syncRecentMenu();
-
-    // we can be sure there is at least 2 itens in ComboBox, "Recent" and the last entered host
-    mRecentComboBox->setCurrentIndex(1);
 }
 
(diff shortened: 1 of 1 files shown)


cipherboy:
@mpreisler  I think this needs a kick.

mpreisler:
ah, yeah

@openscap-ci test this please

matusmarhefka:
@cipherboy On which version have you encountered this?

I cannot reproduce on `1.1.6` (to be more specific I have `scap-workbench-1.1.6-1.fc27.x86_64`). The "User and host" input form is not cleared when I press scan button nor after the scan finishes.

cipherboy:
@mpreisler / @matusmarhefka I can no longer reproduce this. I'm guessing it was a weird artifact of merging v1-1 and master and building when handling merge conflicts. ~shrugs~? I have a single binary which reproduces this issue. Debugging shows that line 95 (of the negative diff) is not getting reached/called. I have not been able to reproduce this under a `git bisect`, though admittedly the tree is a little weird for bisecting. I'm closing this.

Issue #169, opened 2018-06-06
Open other content closes application


cipherboy:
When choosing other content to open, if the other content is not a valid input file (e.g., a JPG), the user will be prompted to select another file. If they then press "cancel" in the file selection dialogue, it will close the application, with the message:

```
14:59:49 | info     | Closing the main window...
```

For better UX, the window should stay open with the last-valid profile.

cipherboy:
Fixed by PR #176 / ba141639569a73aa9a3b18d6704fd79e19c93d85

matejak:
I still experience the issue (on Fedora 34), on a codebase that contains the referenced PR. @jan-cerny could you please check out whether the fix works for you or not any more?

PR #176, opened 2018-06-08
Prevent application from closing when user cancels file selection

cipherboy:
Closes: #169 

Also clarifies that closing the file will lose the tailoring.

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -402,6 +402,9 @@ void MainWindow::openFileDialog()
     if (!QFileInfo(defaultDirectory).isDir())
         defaultDirectory = "";
 
+    // Save the current file path to restore if we're unable to open a valid file
+    const QString currentFile = getOpenedFilePath();
+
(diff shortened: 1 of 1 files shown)


src/MainWindow.cpp
```diff
@@ -402,6 +402,9 @@ void MainWindow::openFileDialog()
     if (!QFileInfo(defaultDirectory).isDir())
         defaultDirectory = "";
 
+    // Save the current file path to restore if we're unable to open a valid file
+    QString current_file = getOpenedFilePath();
```
mpreisler: code style, camelCase please

mpreisler: also, make this const

cipherboy:
@mpreisler Updated! :)

cipherboy:
@mpreisler Updated! :)

mpreisler:
Took 7.5 sec on fedora <-- ccache rules

Issue #134, opened 2017-07-17
Add way to reload datastream


dahaic:
When developing something, I often rebuild datastream and then want to reload it in SCAP Workbench. Currently, there is no option to just reload, so I need to click through directory tree every time this happens.

It would be useful to have such option (for example in ```File/Reload Content```)

It would be great to have also customization preserved.

WesleyCeraso:
I think it should check for changes in the file each X seconds plus when the mainwindow gets focus and just before the scanning. If the file changed it opens a message box asking if the user wants to reload the file or use the old version (it would need some work).

I'm not sure about the customization, I don't know if the scap lib can handle the situation where a customized rule no longer exists in the new version of the file.

dahaic:
I would prefer to have an option there, not the automatic reload, as that would create too much complexity for little benefit.

For the customization - let's use default behavior of SCAP Workbench (whatever it is). I just wanted to avoid situation that it drops customization and I have to create it again, which is painful.

mpreisler:
@WesleyCeraso We can use https://doc.qt.io/qt-4.8/qfilesystemwatcher.html which is much better than any polling solution.

mpreisler:
IMO it should always ask whether you want to reload because it may be dangerous. But we can automatically figure out that the file changed and offer a choice to the user.

cipherboy:
Fixed by #177.

Issue #26, opened 2015-07-07
show notification that opened file was changed


dahaic:
Button `Scan` scans system based on current version of the file on disk - that I understand is a feature. But user may be confused by that, as it may actually scan something different than what is presented by Workbench in the moment of clicking on the `Scan` button.

Would it be possible to show some kind of notification that source file was altered, so user could reload it? [Or at least passively acknowledge it]

cipherboy:
@mpreisler My thoughts on this and #134 are to first copy the scan file to a temporary directory upon loading; `openscap` will then run against that version. When the application is closed, it can remove all temporary files.

We can then use the [FS Watcher](http://doc.qt.io/qt-5/qfilesystemwatcher.html) on the original file and notify of it changing and prompt to reload it.

cipherboy:
Fixed by #177.

PR #177, opened 2018-06-12
Better UX around opened content

cipherboy:
- Implements copying input + dependencies to a temporary directory
- Adds a reload opened content button to the UI (under File)
- Opens an information dialog when the file changes on disk (once)

### DIFF
--- a/include/ScanningSession.h
+++ b/include/ScanningSession.h
@@ -291,14 +291,14 @@ class ScanningSession
         /// Our own tailoring that may or may not initially be loaded from a file
         mutable struct xccdf_tailoring* mTailoring;
 
-        /// Temporary copy of opened XCCDF file
-        QTemporaryDir* mOpenDir;
-        /// Path to temporary XCCDF file
-        QString mOpenPath;
-        /// Path to original XCCDF file
-        QString mOriginalPath;
-        /// Closure of original XCCDF file
-        QSet<QString> mClosureOfFile;
+        /// Temporary copy of opened DS or XCCDF file
+        QTemporaryDir* mTempOpenDir;
+        /// Path to temporary DS or XCCDF file
+        QString mTempOpenPath;
+        /// Path to original DS or XCCDF file
+        QString mOriginalOpenPath;
+        /// Closure of original DS or XCCDF file
+        QSet<QString> mClosureOfOriginalFile;
 
         /// Temporary file provides auto deletion and a valid temp file path
         QTemporaryFile mTailoringFile;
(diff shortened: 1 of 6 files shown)


mpreisler:
#134 #26

src/MainWindow.cpp
```diff
@@ -1507,6 +1552,19 @@ bool MainWindow::unsavedTailoringChanges() const
...
+        return;
+    mFSLastSeen = path;
+
+    QMessageBox::information(
+        this, QObject::tr("SCAP Workbench"),
+        QObject::tr("Opened file was modified after opening: %1\n\n"
+                    "To reload, click File -> Reload Opened Content").arg(path)
+    );
+}
+
```
matejak: At this moment, you show an info dialog. I suggest showing a dialog that would contain `Reload` and `Dismiss/Cancel` buttons, so users can reload right away.

include/ScanningSession.h
```diff
@@ -280,6 +291,15 @@ class ScanningSession
         /// Our own tailoring that may or may not initially be loaded from a file
         mutable struct xccdf_tailoring* mTailoring;
 
+        /// Temporary copy of opened XCCDF file
+        QTemporaryDir* mOpenDir;
+        /// Path to temporary XCCDF file
+        QString mOpenPath;
```
matejak: What about renaming this to `pathToOpenInput` or to something longer, so you can drop the comment? Long names tell you what is the variable content without requiring you to search for the information provided in comments.

src/ScanningSession.cpp
```diff
@@ -71,12 +77,131 @@ struct xccdf_session* ScanningSession::getXCCDFSession() const
...
+     * but this is cleaned on reboot or manually by the user.
+     */
+    if (mOpenDir != NULL) {
+        mOpenDir->remove();
+        delete mOpenDir;
+        mOpenDir = NULL;
+    }
+}
+
+void ScanningSession::cloneToTemporaryFile(const QString& path)
```
matejak: I think that the last `for` loop can be extracted to a private method, as this method is quite complex. It is a pity that we don't have unit tests in the project, because the functionality here would be an ideal candidate.

cipherboy:
@matejak Thanks! I updated the PR.

matejak:
Great work, I like improvements of the dialog and those renames!
However, there is still that issue when you have a file opened and you try to open something else, and you get a message that the current file is a temp file, although you have picked it e.g. from your home directory. After consulting our QE elder wise man @dahaic, the issue is a regression which should be fixed before accepting this PR - it can confuse users, and similar issues like this were fixed in the past.

cipherboy:
@matejak My apologies -- missed that in the first round. :) Updated

matejak:
Wonderful, that was fast! Thank you very much for your contribution!

PR #178, opened 2018-06-12
Style nits

cipherboy:
Paying penance for introducing one in #173 / ae8e3b570e411f494d478acc78705ce66ee7aedf, this fixes the remaining style nitpicks in master.


mpreisler:
Hmm, this is cool but it will make merging from v1-1 to master a PITA. I am wondering if we can apply at least some of this on v1-1.

cipherboy:
@mpreisler Rebased this against `master`.

PR #181, opened 2018-06-25
Fix building issues when using MINGW

redhatrises:
- Add build folder to bring in line with other projects


redhatrises:
Closing as this affects OSX as well.

PR #184, opened 2018-06-29
Add TRAVIS CI for OSX DMG BUILDS

redhatrises:
- Build DMG files using Travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,36 @@
+os: osx
+osx_image: xcode8.3
+sudo: required
+
+language: cpp
+
+compiler:
+  - gcc
+
+before_install:
+  - brew update
+  - brew install jq
+  - brew install cartr/qt4/qt@4
+  - brew install asciidoctor
+
+before_script:
+  - git clone --depth 1 https://github.com/openscap/openscap.git -b master
+  - pushd openscap/build
+  - cmake -DENABLE_PROBES=FALSE ../
+  - make -j 4
+  - make install
+  - popd
+
+script:
+  - ./build-for-osx.sh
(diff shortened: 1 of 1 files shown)


redhatrises:
If we can get an OATH token, the dmg file should automatically be added to the release when a new tag has been pushed.

matejak:
Great PR! Just FYI: You can provide the Github API token via a" secret" env variable in Travis: https://travis-ci.org/OpenSCAP/scap-security-guide/settings, just make sure that the "display value in build log" is off.

redhatrises:
Thanks @matejak! Is there a user or service account that can own the API token? Also, I don't know if I have permissions to setup these things.

matejak:
I did some research and I have found out that you have to use a personal token with scope `public_repo` that you can set up here: https://github.com/settings/tokens
I see that you have set up a Travis check for your `scap-workbench` fork, on https://travis-ci.org/redhatrises/scap-workbench and you already seem to use the `GITHUB_TOKEN` there, how does it work?

redhatrises:
Thanks @matejak. I created my token with repo:status, repo_deployment, and public_repo permissions, and it seems to work great. But what account should own the token?

mpreisler:
Could we please have the job in the OpenSCAP org? https://travis-ci.org/OpenSCAP

redhatrises:
> Could we please have the job in the OpenSCAP org? https://travis-ci.org/OpenSCAP

@mpreisler yes, that's the plan. Just need an account from the OpenSCAP org to create the token.

matejak:
@redhatrises I think that we can't use some kind of service account for the token owner - it has to be one of us. Next, I think that `public_repo` scope alone should be enough, but I may be wrong.
@mpreisler I think that it is not possible to start running Travis jobs in the OpenSCAP org if there is no `.travis.yml` in the project's root in the OpenSCAP project (i.e. not in the fork). I am in favor to merge this if it works reasonably well for Gabe, we can then dispatch a follow-up PR, and Travis may be helpful in ironing that one out.

matejak:
Great, I have set up the github token for OpenSCAP org build of scap-workbench. If this PR won't work for some reason, mistakes can be fixed in a follow-up PR.
Good job, having Mac OS builds delivered automatically will lift some burden from us, and keep general public satisfied at the same time.

PR #185, opened 2018-07-02
Add build directory

redhatrises:


cipherboy:
I've been wanting to add it for the consistency with the other projects,
but thought adding an empty file was too little for a PR :)

PR #188, opened 2018-07-12
Use windows headers instead of unistd.h for Windows builds

redhatrises:


redhatrises:
Travis failed because it hit the GitHub API connection limit... not sure why considering it is using or should be using a GH token.

src/DiagnosticsDialog.cpp
```diff
@@ -26,7 +26,13 @@
 #include <QClipboard>
 
 #include <iostream>
+#if defined(_WIN32)
+#include <io.h>
+#include <windows.h>
+#define usleep Sleep
```
mpreisler: please use the Qt API instead to sleep

redhatrises: @mpreisler is that a note to yourself? https://github.com/OpenSCAP/scap-workbench/commit/4bbc842009d64f627c2f72f59ad553594a4ee41a ;)

src/DiagnosticsDialog.cpp
```diff
@@ -51,6 +51,14 @@ DiagnosticsDialog::DiagnosticsDialog(QWidget* parent):
     dumpVersionInfo();
 }
 
+class Sleeper : public QThread
+{
+public:
+    static void usleep(unsigned long usecs){QThread::usleep(usecs);}
+    static void msleep(unsigned long msecs){QThread::msleep(msecs);}
+    static void sleep(unsigned long secs){QThread::sleep(secs);}
+};
```
mpreisler: what is this abomination?!

Why not just use `QThread::sleep` directly? 

PS: interval is an int, times 1000 clearly means we want to sleep a certain number of seconds...

src/DiagnosticsDialog.cpp
```diff
@@ -51,6 +51,12 @@ DiagnosticsDialog::DiagnosticsDialog(QWidget* parent):
     dumpVersionInfo();
 }
 
+class Sleeper : public QThread
```
mpreisler: please add a comment so that the next person reading this can figure out why it's even here...

src/DiagnosticsDialog.cpp
```diff
@@ -51,6 +51,12 @@ DiagnosticsDialog::DiagnosticsDialog(QWidget* parent):
...
+{
+public:
+    static void usleep(unsigned long usecs){QThread::usleep(usecs);}
```
mpreisler: I suggest using `sleep` instead

src/DiagnosticsDialog.cpp
```diff
@@ -64,7 +74,7 @@ void DiagnosticsDialog::waitUntilHidden(unsigned int interval)
     while (isVisible())
     {
         QAbstractEventDispatcher::instance(0)->processEvents(QEventLoop::AllEvents);
-        usleep(interval * 1000);
+        Sleeper::sleep(interval * 1000);
```
mpreisler: I am 100% confident you didn't test this :-D

redhatrises:
Closing as QT5 is now used instead of QT4.

PR #190, opened 2018-07-18
Use QT5 for travis builds

redhatrises:


mpreisler:
let's just do this for v1-2 and then it will get merged up, no need for this PR to stay open

PR #191, opened 2018-07-18
[v1-2] Fix OSX builds

redhatrises:


redhatrises:
@mpreisler I don't suppose that we can get rid of `0_alpha2` in OpenSCAP CMake version can we? See https://travis-ci.org/OpenSCAP/scap-workbench/builds/405559444#L2882

jan-cerny:
@mpreisler @redhatrises Now it looks like a bad decision to introduce "_alphaX" suffix to the version string in OpenSCAP.  It has already caused problems with RPM builds, which were solved by using underscore instead of originally proposed dash. Moreover, it has already caused problems with building Windows installer, because MSI generator requires only integers there. (See discussion in PR https://github.com/OpenSCAP/openscap/pull/1116) Now it causes problems with SCAP Workbench.

A workaround could be to change this condition in SCAP Workbench in a way that it first tests if the version is greater than or equal 1.3 (without testing the third number) and if it isn't it would test if the version is greater than or equal 1.2.12. I haven't checked if that works, though.

mpreisler:
I've seen other projects use some very high patch number for this. Like 1.2.9999

redhatrises:
@jan-cerny @mpreisler why all of the sudden are we adding alpha versioning to OpenSCAP? What purpose is it serving when we aren't branching the versions? Seems like a lot of effort for very little gain? 

If we have to do alpha or dev versions, I would +1 adding `.99` instead of `_alpha` as I have seen other projects do the same.

redhatrises:
@mpreisler should we also flip the primary branch to `v1-2` instead of `v1-1`?

mpreisler:
> @mpreisler should we also flip the primary branch to v1-2 instead of v1-1?

yup, done, thx!

jan-cerny:
@redhatrises The purpose of that was to avoid need to support the API because we will most likely change the API before final 1.3.0 is released.

I like the idea with 9999 though.
@mpreisler What can we do if we already started with the `alpha` suffix?

.travis.yml
```diff
@@ -8,7 +8,7 @@ compiler:
   - gcc
 
 env:
-  - PATH="/usr/local/opt/qt:$PATH"
+  - PATH="/usr/local/opt/qt:/usr/local/opt/qt/bin:$PATH"
```
mpreisler: do we really need both opt/qt AND opt/qt/bin?

CMakeLists.txt
```diff
@@ -117,6 +117,27 @@ set(SCAP_WORKBENCH_LINK_LIBRARIES
     Qt5::Widgets Qt5::XmlPatterns
     ${OPENSCAP_LIBRARIES})
 
+# ---------- RPATHS for linking
+
+# see https://cmake.org/Wiki/CMake_RPATH_handling
```
mpreisler: outdated link

CMakeLists.txt
```diff
@@ -117,6 +117,27 @@ set(SCAP_WORKBENCH_LINK_LIBRARIES
...
+
+# use, i.e. don't skip the full RPATH for the build tree
+set(CMAKE_SKIP_BUILD_RPATH FALSE)
```
mpreisler: this is the default, why are we setting it to the same value?

CMakeLists.txt
redhatrises: @mpreisler fixed

CMakeLists.txt
redhatrises: @mpreisler fixed

.travis.yml
redhatrises: @mpreisler fixed

CMakeLists.txt
redhatrises: @mpreisler went ahead and only did rpath handling for Apple as I am not sure if we need to handle it for Linux and windows.

CMakeLists.txt
```diff
@@ -125,6 +125,26 @@ set(SCAP_WORKBENCH_LINK_LIBRARIES
...
+if (APPLE)
+    # when building, don't use the install RPATH already
+    # (but later on when installing)
+    set(CMAKE_BUILD_WITH_INSTALL_RPATH ON)
+
+    set(CMAKE_INSTALL_RPATH ${CMAKE_INSTALL_FULL_LIBDIR})
+
+    # add the automatically determined parts of the RPATH
+    # which point to directories outside the build tree to the install RPATH
+    set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
```
yuumasato: Why the need for this?
I guess all libraries needed are in build tree, or default path of system.

CMakeLists.txt
```diff
@@ -125,6 +125,26 @@ set(SCAP_WORKBENCH_LINK_LIBRARIES
...
+# ---------- RPATHS for linking
+
+# see https://gitlab.kitware.com/cmake/community/wikis/doc/cmake/RPATH-handling
+
+if (APPLE)
+    # when building, don't use the install RPATH already
+    # (but later on when installing)
+    set(CMAKE_BUILD_WITH_INSTALL_RPATH ON)
+
+    set(CMAKE_INSTALL_RPATH ${CMAKE_INSTALL_FULL_LIBDIR})
```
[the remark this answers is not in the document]
redhatrises: @yuumasato yes... specifically `libopenscap`

CMakeLists.txt
redhatrises: @yuumasato this is needed so that we can use `libopenscap` and place it in `./scap-workbench.app/Contents/Frameworks/`.

CMakeLists.txt
```diff
@@ -125,6 +125,26 @@ set(SCAP_WORKBENCH_LINK_LIBRARIES
...
     ${OPENSCAP_LIBRARIES})
 
+# ---------- RPATHS for linking
+
+# see https://gitlab.kitware.com/cmake/community/wikis/doc/cmake/RPATH-handling
+
+if (APPLE)
+    # when building, don't use the install RPATH already
+    # (but later on when installing)
+    set(CMAKE_BUILD_WITH_INSTALL_RPATH ON)
```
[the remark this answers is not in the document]
redhatrises: @yuumasato you are right! Fixed.

CMakeLists.txt
```diff
@@ -125,6 +125,26 @@ set(SCAP_WORKBENCH_LINK_LIBRARIES
...
+
+    set(CMAKE_INSTALL_RPATH ${CMAKE_INSTALL_FULL_LIBDIR})
+
+    # add the automatically determined parts of the RPATH
+    # which point to directories outside the build tree to the install RPATH
+    set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
+
+    # Turn on RPATH for OSX for policy warning
+    set(CMAKE_MACOSX_RPATH ON)
+    # ---------- CONFIGURATION
```
mpreisler: this comment is probably misplaced

CMakeLists.txt
```diff
@@ -125,6 +125,26 @@ set(SCAP_WORKBENCH_LINK_LIBRARIES
     Qt5::Widgets Qt5::XmlPatterns
     ${OPENSCAP_LIBRARIES})
 
+# ---------- RPATHS for linking
+
+# see https://gitlab.kitware.com/cmake/community/wikis/doc/cmake/RPATH-handling
+
+if (APPLE)
```
mpreisler: let's make this universal and not apple only please

CMakeLists.txt
yuumasato: I think `libopenscap` will be found because `CMAKE_INSTALL_RPATH` was already set to `${CMAKE_INSTALL_FULL_LIBDIR}`.

CMakeLists.txt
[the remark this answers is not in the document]
yuumasato: The second line of comment is also strange.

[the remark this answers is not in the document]
yuumasato: Also, during build time, we haven't yet copied `libopenscap` to be bundled in workbench. And it was installed in the system defaults during `before-script`, so it should be ok, no?

CMakeLists.txt
redhatrises: @mpreisler done

mpreisler: wait, have you reworked and tested it or just removed the if?

redhatrises: @mpreisler unless you are seeing something different, compiled and ran it locally and appears to be working.

yuumasato:
@openscap-ci test this please

jan-cerny:
Travis CI fails on installing the user manual OpenSCAP user's manual. It can be either fixed in OpenSCAP or we can use `-DENABLE_DOCS=FALSE` in the SCAP Workbench build.

jan-cerny:
The failure is due to move of SCAP Security Guide to Compliance as Code content.

yuumasato:
> The failure is due to move of SCAP Security Guide to Compliance as Code content.

PR #209 attempts to fix

yuumasato:
@redhatrises I have tried updating the `GITHUB_TOKEN`, but it doesn't seem to be used in the `curl` call.

redhatrises:
@yuumasato the `GITHUB_TOKEN` is just so the api call doesn't hit the github api call limit. I assume that the `GITHUB_TOKEN` has access to `https://github.com/ComplianceAsCode/content` with the `public_repo` permission?

yuumasato:
@redhatrises yes, locally the token works for me.

yuumasato:
It seems that Travis doesn't want to use the `GITHUB_TOKEN` I have configured.
In GH, when listing access token, it says "Never used".

yuumasato:
@openscap-ci test this please

yuumasato:
Closing and re-opening to trigger new submission to Travis.

yuumasato:
My conclusion is that Travis CI for PRs never worked, it only worked after Merge commits, and on schedule basis.
The use of encrypted variable on  `$GITHUB_TOKEN` triggers a clean up of i on PRs, check https://docs.travis-ci.com/user/pull-requests/#pull-requests-and-security-restrictions. 

I thought it worked in the past, but fro https://travis-ci.org/OpenSCAP/scap-workbench/pull_requests I see that it never did.
The only situation in which it worked, was a PR from the same fork, in #201.

CMakeLists.txt
```diff
@@ -125,6 +125,23 @@ set(SCAP_WORKBENCH_LINK_LIBRARIES
     Qt5::Widgets Qt5::XmlPatterns
     ${OPENSCAP_LIBRARIES})
 
+# ---------- RPATHS for linking
+
+# see https://gitlab.kitware.com/cmake/community/wikis/doc/cmake/RPATH-handling
+
+# when building, use the install RPATH
+# (but later on when installing)
+set(CMAKE_BUILD_WITH_INSTALL_RPATH ON)
```
yuumasato: I'm no sure we need to set `RPATH` on Linux and Windows.
I don't see any harm, but no benefit either.

redhatrises:
@yuumasato I have added a workaround that doesn't use the github api. It's not as elegant, but it works. We will of course still need the secure token for pushing a release.

CMakeLists.txt
yuumasato: I realized this is needed on Apple because we don't run `install` target, we just build and run a script to generate the `dmg`.
I would add a comment and make this Apple only, and let CMake re-link on `install` target on Linux.

CMakeLists.txt
```diff
@@ -125,6 +125,23 @@ set(SCAP_WORKBENCH_LINK_LIBRARIES
...
+
+set(CMAKE_INSTALL_RPATH ${CMAKE_INSTALL_FULL_LIBDIR})
```
yuumasato: Doesn't `CMAKE_INSTALL_FULL_LIBDIR` resolve into a system path? Why add it to RPATH?

CMakeLists.txt
yuumasato: But even then, at build time, openscap libraries are in system path, https://travis-ci.org/OpenSCAP/scap-workbench/jobs/481474332#L2433.
And the `dmg` is generated with openscap libraries at `./scap-workbench.app/Contents/Frameworks/`

redhatrises:
> > @yuumasato I have added a workaround that doesn't use the github api. It's not as elegant, but it works. We will of course still need the secure token for pushing a release.
> 
> The workaround looks ok. I believe the release process should work fine, as it will trigger from tags, and not from PRs.
> 
> @redhatrises I'm not convinced that `RPATH` changes are needed.
> Do you mind moving them to separate PR?
> And them we can merge this PR with Qt5 and ComplianceAsCode fixes

@yuumasato IIRC, CMake policy complains on OSX when RPATH is not set. IIRC, without setting rpath, scap-workbench refuses to run. Also, all the cmake rpath settings match what we have set in oscap. See https://github.com/OpenSCAP/openscap/pull/1084. Originally, I had just RPATH only for APPLE, but it was asked to have rpath turned on universally. I feel like we should be consistent here.

yuumasato:
The [policy](https://cmake.org/cmake/help/v3.0/policy/CMP0042.html) would like `(CMAKE_|)MACOSX_RPATH` to be true, but I still don't understand how the current`RPATH`configuration helps with OSX build and run.

Ack, as it seems that these changes won't cause trouble.

PR #193, opened 2018-07-19
pkexec: avoid potential local root exploit by using PKEXEC_UID and sudo

mgerstner:
If an admin relaxes the required polkit authentication for running
scap-workbench-oscap.sh from auth_admin to auth_self or yes, then the
current implementation of the wrapper script allows for a local root
exploit.

A command line like this would overwrite /etc/shadow with a file owned
by the non-privileged user:

pkexec --disable-internal-agent /usr/lib64/scap-workbench/scap-workbench-oscap.sh 1000 100 \
	xccdf eval --profile Default --oval-results --results /etc/shadow \
	--results-arf /tmp/scap.results.arf --report /tmp/scap.report \
	--progress /usr/share/openscap/scap-yast2sec-xccdf.xml

The copying of the target files needs to be done in the context of the
unprivileged user to prevent any symlink attacks or maliciously
specified paths. This is done by using sudo as a frontend to cp.

Also the user should not pass his own uid and gid. This would allow to
change ownership of files to arbitrary other users. Instead pkexec
offers the PKEXEC_UID environment variable which contains the uid of the
authenticated user. The gid can be derived from the uid.

### DIFF
--- a/scap-workbench-oscap.sh
+++ b/scap-workbench-oscap.sh
@@ -25,13 +25,14 @@ trap "" SIGHUP SIGINT
 # valuable gets skipped
 echo "Dummy text" 1>&2
 
-wrapper_uid=$1
-shift
-wrapper_gid=$1
-shift
+# prevent world-readable files being created
+umask 0007
 
 real_uid=`id -u`
 real_gid=`id -g`
+
+wrapper_uid=${PKEXEC_UID:-${real_uid}}
+wrapper_gid=$(id -g ${wrapper_uid})
 
 TEMP_DIR=`mktemp -d`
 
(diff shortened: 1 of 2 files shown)


cipherboy:
@mgerstner -- Thanks for the patch. We'll hope to review it soon. :)

scap-workbench-oscap.sh
```diff
@@ -84,19 +85,34 @@ RET=$?
...
+    do_chown=true
+fi
+
 function chown_copy
 {
     local what="$1"
     local where="$2"
 
-    [ ! -f "$what" ] || cp "$what" "$where"
+    [ ! -f "$what" ] && return
```
matejak: I would slightly prefer the `[ -f "$what" ] || return` form of this statement, as it is required for running this script with `set -e`.

mgerstner:
Thanks @matejak for reviewing. I adjusted the PR accordingly to incorporate your suggestions.

matejak:
Great job, thank you!
Note for future: The Travis test failiure has nothing to do with this PR.

cipherboy:
Yes, thank you @mgerstner!

PR #195, opened 2018-07-19
Use QT interals for sleep rather than unistd.h

redhatrises:


src/DiagnosticsDialog.cpp
```diff
@@ -64,7 +64,7 @@ void DiagnosticsDialog::waitUntilHidden(unsigned int interval)
     while (isVisible())
     {
         QAbstractEventDispatcher::instance(0)->processEvents(QEventLoop::AllEvents);
-        usleep(interval * 1000);
```
jan-cerny: `usleep` takes microseconds, `QThread::sleep` takes seconds, but you are passing there miliseconds, which means it will sleep for a 1000 times longer period.
See http://doc.qt.io/qt-5/qthread.html

redhatrises: So wait.... @mpreisler wanted to use sleep over usleep, but maybe, I got confused as to what @mpreisler wanted?

jan-cerny: Take a look at https://github.com/OpenSCAP/scap-workbench/blob/v1-2/include/DiagnosticsDialog.h#L87. The default value for `interval` is 100. That means it slept for 0.1 second, but with your patch it will sleep for 100 seconds. I doubt that we want that.

redhatrises: Not sure what I am confused about, but anyway, kept the same logic, just using QT usleep.

mpreisler: *insert cheeky joke about Americans and SI units*

PR #200, opened 2018-07-25
Use AsciiDoc instead of Asciidoctor

jan-cerny:
It is needed to use AsciiDoc instead of AsciiDoctor to be able to build the User Manual on some Linux distributions. We should use AsciiDoc only to make sure the resulting document is always the same.

Unfortunately, AsciiDoc is not able to correctly set the MIME data for base64 encoded SCAP Workbench SVG logo. It defines there `data:image/svg;` instead of correct `data:image/svg+xml;`. Therefore, I have replaced the SVG logo by a PNG logo.


cipherboy:
I confirmed this build failed without `asciidoc` installed. Then I validated that it built with `asciidoc` installed, with no errors. Then I loaded the output file (`/home/ascheel/GitHub/jan-cerny/scap-workbench/doc/user_manual.html`) and saw a pretty SCAP Workbench logo and documentation and verified that it seemed to look correct. 


Looks good, thanks @jan-cerny! 

(Travis failures look to still be a problem with OS X and unrelated to this PR)

PR #201, opened 2018-08-09
Support oscap version suffix when scanning remote

mpreisler:
Previously we only supported versions such as 1.2.3, now we also support
1.2.3_alpha4.


cipherboy:
ACK.

PR #221, opened 2018-12-14
Sort content labels lexicographically and allow setting of default product

matejak:
That way, the combo box contents are going to be consistent and e.g. RHEL8 will always come after RHEL7 and so on.


jan-cerny:
Hi @matejak , what remains to be done here?

matejak:
The problem is that this doesn't solve the primary UX problem, which is that we would like to offer RHEL8 option by default on RHEL8 systems, and RHEL7 on RHEL7 systems and so on. However, ordering may make sense by itself, in which case it can be merged.

jan-cerny:
@matejak please rebase

matejak:
The default item may now be controlled using a cmake argument `SCAP_WORKBENCH_PREFERRED_DATASTREAM_BASENAME`.

jan-cerny:
@matejak Thanks for extending this. However, the default item setting works only sometimes. I suspect it conflicts with the `favorite` variable. If I set SCAP_WORKBENCH_PREFERRED_DATASTREAM_BASENAME to ssg-rhel7-ds.xml then RHEL7 is the default. But if I set it to ssg-ubuntu1804-ds.xml, Fedora is default, and Ubuntu 18.04 is only moved a little up, right after RHEL8.

I think that marking the default item as favourite would help us. At the same time I suggest dropping the favourite mechanism completely and only provide a lexicographically sorted list.

matejak:
Good idea, the favourite concept shouldn't be defined in the sourcecode.
The new behavior allows specification of comma-separated list of datastream basenames, which allows compile-time composition analogy of the Favourites section, plus the ability to define order within the section.

CMakeLists.txt
```diff
@@ -332,6 +332,8 @@ if (CPPCHECK_EXECUTABLE)
     )
 endif()
 
+set(SCAP_WORKBENCH_PREFERRED_DATASTREAM_BASENAMES "ssg-<productname1>-ds.xml" CACHE STRING "Comma-separated list of datastream that should be the first ones on the selection list in the initial dialogue.")
```
jan-cerny:
```suggestion
set(SCAP_WORKBENCH_PREFERRED_DATASTREAM_BASENAMES "ssg-<productname1>-ds.xml" CACHE STRING "Comma-separated list of datastreams that should be the first ones on the selection list in the initial dialogue.")
```

PR #227, opened 2019-07-10
Master - creation of audits with scap workbench from roles created in Ansible

sapariciol:
am creating profiles in scap-workbench from roles made in Ansible, but there are compatibility errors, I want to know if I can create audits from the roles created in Ansible?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,10 +7,13 @@ language: cpp
 compiler:
   - gcc
 
+env:
+  - PATH="/usr/local/opt/qt/bin:$PATH"
+
 before_install:
   - brew update
   - brew install jq
-  - brew install cartr/qt4/qt@4
+  - brew install qt5
   - brew install asciidoc
 
 before_script:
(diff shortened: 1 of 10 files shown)


ggbecker:
Hi @sapariciol, there is no easy way to convert your ansible roles to an `audit` profile. You need to understand what your ansible role do and identify which rule in the [ComplianceAsCode/content](https://github.com/ComplianceAsCode/content/) covers that.

The ComplianceAsCode/content is the project which produces content consumed by OpenSCAP.

For example: to check some `sshd` configuration we can use this rule: [SSHD Enable StrictModes](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/services/ssh/ssh_server/sshd_enable_strictmodes/rule.yml).

So first you need to identify the mapping of your ansible roles to rules provided by ComplianceAsCode/content, then you can use SCAP-Workbench to taylor the profile and select these mapped rules to craft a profile representing your ansible roles.

This is some information you might wanna know, but I would need more info on what exactly you are trying to achieve.

sapariciol:
I would like to have a customized profile in SCAP-Workbench with websites that my company needs in the network security part,

PR #230, opened 2019-08-14
[v1-1] Remove library so versioning for OSX builds

### LINKED ISSUE #202: lib change - test?
Due to OpenSCAP/openscap#1158, the library so number will be changing. At minimum, we should update the OSX build:

https://github.com/OpenSCAP/scap-workbench/blob/v1-2/build-for-osx.sh#L7

(v1-2, which will get merged into master).

But we should also test that it builds with a new version of `scap-workbench`.

redhatrises:
- Fixes #202

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,8 +23,17 @@ before_script:
 
 script:
   - ./build-for-osx.sh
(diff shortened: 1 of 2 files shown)


yuumasato:
LGTM

redhatrises:
Thanks for the review and merge @yuumasato!

PR #232, opened 2019-08-23
[v1-2] Fix results font color

redhatrises:
- Sets the Rule title and description font color to black. The font became hard to read when using dark themes in Gnome


redhatrises:
Bump for review.

shawndwells:
The test are failing on MacOS because QT has deprecated some of the functions in use. Ignoring - some elements need to be updated for QT5.

Issue #196, opened 2018-07-23
Problems with temporary directories


jan-cerny:
Version: scap-workbench-1.2.0-1.fc27.x86_64

See the Window that appeared after finished scanning Standard Fedora profile.
![screenshot from 2018-07-23 09-13-29](https://user-images.githubusercontent.com/9050916/43062350-ceac15f8-8e58-11e8-8dc6-2aa54ec939e8.png)

mpreisler:
looks like a word splitting fail

jan-cerny:
I have verified that this particular error has already been fixed in upstream by https://github.com/OpenSCAP/scap-workbench/pull/213.

However, there are different warning messages that pop up in this window. They haven't been fixed yet, but they have already been reported in https://github.com/OpenSCAP/scap-workbench/issues/187 and https://bugzilla.redhat.com/show_bug.cgi?id=1781074

PR #239, opened 2019-12-13
Remove space from temporary file names

jan-cerny:
We use QTemporaryFile to create temporary files. According to
https://doc.qt.io/qt-5/qtemporaryfile.html, the default filename
is determined from QCoreApplication::applicationName(). It returns
"SCAP Workbench" which leads to creating temporary files with a space
character in their name, eg. "/tmp/SCAP Workbench.XM8663". The space
can cause problems with missing quotes s.a. the problems described in
https://github.com/OpenSCAP/scap-workbench/issues/196.
To avoid these problems, it's probably better to create the temporary
files without the space. I think I can use function setFileTemplate
for this purpose.


jan-cerny:
Replaced by https://github.com/OpenSCAP/scap-workbench/pull/240

matejak:
This is not a bad idea, although I would go for e.g. `SpacelessQTemporaryFile` declared in `Utils.h` that would extend the `QTemporaryFile` just by calling the `setFileTemplate` method. Then, if the code uses this class instead of the stock one, there won't be any space.

jan-cerny:
I have changed it using @matejak's suggestion. I have noticed that we use our own version of class for creating temporary directories. That was needed when we used Qt 4, because Qt 4 didn't have QTemporaryDir class. But since SCAP Workbench release 1.2.0 we support only Qt 5, where this class is present, therefore we can drop the custom implementation and just use the QTemporaryDir from Qt 5 library.

src/Utils.cpp
```diff
@@ -217,3 +217,9 @@ const QString& getSetSidPath()
     return ret;
 #endif
 }
+
+SpacelessQTemporaryFile::SpacelessQTemporaryFile () :
+    QTemporaryFile(QDir::cleanPath(QDir::tempPath() + "/" + "XXXXXX")) {}
+
+SpacelessQTemporaryDir::SpacelessQTemporaryDir () :
+    QTemporaryDir(QDir::cleanPath(QDir::tempPath() + "/" + "XXXXXX")) {}
```
matejak: Could you please add an indication that this is scap-workbench temp file/dir?

jan-cerny:
@matejak I have added scap-workbench to temporary file name template.

matejak:
Great, thanks!

PR #240, opened 2019-12-13
Do not set application name

jan-cerny:
We use QTemporaryFile to create temporary files. According to
https://doc.qt.io/qt-5/qtemporaryfile.html, the default filename is
determined from QCoreApplication::applicationName(). It returns "SCAP
Workbench" which leads to creating temporary files with a space
character in their name, eg. "/tmp/SCAP Workbench.XM8663". The space can
cause problems with missing quotes s.a. the problems described in issue
application name manually. Hopefully it will not cause any problem.  If
not set, the application name defaults to the executable name, which is
`scap-workbench`. It doesn't affect the window title.

### DIFF
--- a/src/Application.cpp
+++ b/src/Application.cpp
@@ -37,7 +37,6 @@ Application::Application(int& argc, char** argv):
     setOrganizationName("SCAP Workbench upstream");
     setOrganizationDomain("https://www.open-scap.org/tools/scap-workbench");
 
-    setApplicationName("SCAP Workbench");
     setApplicationVersion(SCAP_WORKBENCH_VERSION);
 
     mMainWindow = new MainWindow();


evgenyz:
This change will affect the location of the config file. It used to store some settings for the remote connection, the directory and the time of usage of the last tailoring file. Insted of `~/.config/SCAP Workbench upstream` it would be `~/.config/scap-workbench`.

So there are three consequences:
- newer version would not be able to pick up the old settings;
- old settings file and directory would be orphaned;
- the name of the directory would be less "user readable".

Are we OK with this?

evgenyz:
I'll correct myself. The directory name is coming from `OrganizationName` not `ApplicationName`. But still the config file name would be formed from `ApplicationName`. So `/.config/SCAP Workbench upstream/SCAP Workbench.conf` would become `/.config/SCAP Workbench upstream/scap-workbench.conf`.

We might also want to review the configuration naming policy as a whole.

matejak:
What I don't like about this solution is that somebody may re-introduce the application name for another reason. What about [Andreas Bleischwitz's suggesstion](https://bugzilla.redhat.com/show_bug.cgi?id=1781074#c5)?

>The issue may be caused by using the default template of `QTemporaryFile()` which uses `QCoreApplication::applicationName()` suffixed by some arbitrary numbers:
>>`QTemporaryFile::QTemporaryFile()`
>> Constructs a `QTemporaryFile` using as file template the application name returned by `QCoreApplication::applicationName()` (otherwise qt_temp) followed by ".XXXXXX".
>>The file is stored in the system's temporary directory, `QDir::tempPath()`.
>
>In order to avoid those whitespaces `QTemporaryFile()` will need to get a suitable template as parameter or `QTemporaryFile::setFileTemplate()` to be initailized.

jan-cerny:
@matejak That was my original idea, and as you already noticed I have originally proposed that in #240.

jan-cerny:
@evgenyz I completely forgot about the config file. I don't think the config file is a documented feature, but it's nice. However, since spaces in file names require quoting, I would prefer if there are no spaces in the config file name as well.

evgenyz:
@jan-cerny It is documented that `applicationName` property could be used in `QSettings` on certain conditions (https://doc.qt.io/qt-5/qcoreapplication.html#applicationName-prop), it's just not right away clear how it is used.

I don't like it that way too, and I don't like the organization name as well. But we should clean up or recycle old configuration in order to be nice to our users if we are going to proceed, IMHO.

jan-cerny:
The temp files will be resolved by #239. The configuration file can be solved by a separate PR.

PR #231, opened 2019-08-15
Remove deprecated or obsolete QT4 code

redhatrises:
Updates deprecated QT4 code to QT5 which fixes build failures related to deprecated QT components.

PR #241, opened 2020-01-03
Stopped using deprecated escape static method.

matejak:
See also: https://wiki.qt.io/Transition_from_Qt_4.x_to_Qt5#Qt::escape_is_deprecated


redhatrises:
@matejak if you want to add commits to #231, there are additional fixes there that this fix is already in.

matejak:
Thanks for the notice, closing as this is subset of #231.

PR #245, opened 2020-02-06
Sort items in Tailoring window

jan-cerny:
It will display first groups, then values, and then rules.
Each part will be sorted alphabetically according to items titles.

Sorting improves orientation in large groups, eg. SELinux group.

Fixes: RHBZ#1798739


matejak:
Code looks good and works great, thank you very much for the PR!

PR #247, opened 2020-02-12
Fixed integration dialog sizing

matejak:
Before, the dialog/label size was given by static values. This PR makes the dialog to resize depending on the text length, which should be more robust.

- Made the label height auto-adjustable - see https://stackoverflow.com/a/19294869
- Increased horizontal spacing between logo / label
- Removed a "widget" that isn't used, and messed up the layout
- Decreased the default dialog height, that acts as a minimal height.

Fixes: rhbz#1743713


jan-cerny:
Hopefully it improves the situation. Thank you very much.

PR #248, opened 2020-02-25
Generate result-based remediation from tailored profile

jan-cerny:
Users can generate remediation script from scan results
of a tailored profile.

Unfortunately, the current design of SCAP Workbench doesn't allow
a clear way of doing this. The scan is run in a separated oscap
process. SCAP Workbench doesn't have access to oscap internal
xccdf_session which creates the ARF. It can't obtain the Tailoring
component reference ID.

Instead, we will save the tailoring document to a temporary file
and use the temporary file when generating the remediation.

Resolves: RHBZ#1640715

### DIFF
--- a/include/RemediationRoleSaver.h
+++ b/include/RemediationRoleSaver.h
@@ -98,33 +98,35 @@ class PuppetProfileRemediationSaver : public ProfileBasedRemediationSaver
 class ResultBasedProcessRemediationSaver : public RemediationSaverBase
 {
     public:
-        ResultBasedProcessRemediationSaver(QWidget* parentWindow, const QByteArray& arfContents,
+        ResultBasedProcessRemediationSaver(
+                QWidget* parentWindow, const QByteArray& arfContents, const QString& tailoringFilePath,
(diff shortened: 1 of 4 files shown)


jan-cerny:
This leads me to the question of removing the whole `#ifndef SCAP_WORKBENCH_USE_LIBRARY_FOR_RESULT_BASED_REMEDIATION_ROLES_GENERATION` block. It says:  "If enabled, result-based remediation roles will be generated by calls to the libopenscap library (instead of being generated by the oscap subprocess). Requires openscap>=1.2.16". Do we still want to support builds against older versions or can we drop it? What do you think?

matejak:
> Do we still want to support builds against older versions or can we drop it? What do you think?

I think that we can drop the support for scanners such old in upcoming workbench versions.

matejak:
The PR works, the solution with using the tailoring file works. If the user decides to delete the tailoring file (that is located in `/tmp`), generation of remediation fails, and the error mentioning the deleted file appears in the error log, which I see as acceptable.

jan-cerny:
@matejak The removal has been submitted as https://github.com/OpenSCAP/scap-workbench/pull/249

PR #249, opened 2020-03-02
Use only library calls to generate remediation

jan-cerny:
Removes CMake option
SCAP_WORKBENCH_USE_LIBRARY_FOR_RESULT_BASED_REMEDIATION_ROLES_GENERATION
and removes the code that is used when this option is not set.
That means the remediations will be generated using libopenscap
library calls. The removed code executed "oscap" command to
do the same thing.


matejak:
Good riddance - tailored remediation generation still works.

PR #250, opened 2020-03-03
Fix terminology

jan-cerny:
Roles don't exist.


matejak:
I absolutely agree. Nice UX fix, thanks!

PR #259, opened 2020-05-26
Use QT provided macro function to version check - deprecation errors

carlosmmatos:
This allows us to prevent fatal deprecation warnings from qt 5.13 to 5.14. This is also needed to support new CI pipelines that will be using the latest qt5.14.2+.


redhatrises:
@openscap-ci add to whitelist

carlosmmatos:
To add more context - this issue was found when working on porting the travis ci to github actions which uses the latest macOS 10.15 and QT5.14.2. Travis CI defaults currently to macOS 10.13 and QT5.13.2. This change works with both scenarios.

redhatrises:
LGTM

PR #260, opened 2020-05-26
Fix CMakeLists OSX build issue

carlosmmatos:
Can't find openscap library since post macOS 10.13 the standard path has been changed. This fix will use the full path to declare the location of the openscap library


jan-cerny:
Great finding!

PR #261, opened 2020-06-01
Porting CI from Travis to Github actions

carlosmmatos:
New Github Actions file follows similar convention to Travis with a few minor differences.
By default - GHA does not let you use normal syntax to skip jobs when testing. The current
setup allows for a dev to use either [skip ci] or [skip-ci] in order to not kick off a job.
This is similar to other CI tools except that in this instance we have full control of the syntax
we want to check for in order to skip a job.

Other fixes include additional verbosity on the build-for-osx and osx-create-dmg scripts. There
is also a new travis-ci file that is deprecated, but shown in case someone would like to use it
for further testing.


carlosmmatos:
**This PR depends on #260** 

Github doesn't seem to have a decent way to setup PR dependencies - so I'll list it out here.

jan-cerny:
@carlosmmatos I have merged #260 so now you can rebase.

carlosmmatos:
Link to Github Actions CI results - after rebasing.

https://github.com/carlosmmatos/scap-workbench/actions/runs/122805050

jan-cerny:
@carlosmmatos Thank you!

PR #265, opened 2020-08-04
Update list of required packages

jan-cerny:
qt5-devel is only a meta package. The real dependencies are qt5-qtbase-devel and qt5-qtxmlpatterns-devel. qt5-devel is
removed in Fedora 33. See https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/message/WO625MVYEAAJNHNRLEEJDVZTIWMQOBRR/

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ How to run it out of the box
 
 required dependencies:
 ```console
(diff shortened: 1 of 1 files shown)


matejak:
LGTM, thanks!

PR #266, opened 2020-10-07
Replace obsolete QString::SkipEmptyParts

jan-cerny:
Starting from Qt 5.15, the enum QString::SplitBehavior is obsolete
and Qt::SplitBehavior should be used instead.
See:
https://doc.qt.io/qt-5.15/qstring-obsolete.html#SplitBehavior-enum
https://doc.qt.io/qt-5/qt.html#SplitBehaviorFlags-enum


matejak:
LGTM, thanks!

PR #270, opened 2021-01-26
Implement possibility to scan by sudoers.

matejak:
- The remote scanning dialog got a "user is sudoer" checkbox.
- Dry run can make use of sudo in connection with `oscap-ssh`.
- The scanning procedure uses sudo invocation as part of the `ssh` command.

This PR depends on #271 as it makes use of advanced message filtering.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1877522

### DIFF
--- a/include/OscapScannerRemoteSsh.h
+++ b/include/OscapScannerRemoteSsh.h
@@ -36,6 +36,7 @@ class OscapScannerRemoteSsh : public OscapScannerBase
         OscapScannerRemoteSsh();
         virtual ~OscapScannerRemoteSsh();
 
+        void setUserIsSudoer(bool userIsSudoer);
         virtual void setTarget(const QString& target);
         virtual void setSession(ScanningSession* session);
 
@@ -57,6 +58,7 @@ class OscapScannerRemoteSsh : public OscapScannerBase
         void removeRemoteDirectory(const QString& path, const QString& desc);
 
         SshConnection mSshConnection;
+        bool mUserIsSudoer;
 };
 
 #endif
--- a/include/OscapScannerRemoteSsh.h
+++ b/include/OscapScannerRemoteSsh.h
@@ -43,6 +43,11 @@ class OscapScannerRemoteSsh : public OscapScannerBase
         virtual QStringList getCommandLineArgs() const;
         virtual void evaluate();
 
+    protected:
+
+       virtual void selectError(MessageType& kind, const QString& message);
+       virtual void processError(QString& message);
+
     private:
         void ensureConnected();
 
--- a/include/OscapScannerRemoteSsh.h
+++ b/include/OscapScannerRemoteSsh.h
@@ -31,11 +31,12 @@ class OscapScannerRemoteSsh : public OscapScannerBase
     Q_OBJECT
 
     public:
-        static void splitTarget(const QString& in, QString& target, unsigned short& port);
+        static void splitTarget(const QString& in, QString& target, unsigned short& port, bool& userIsSudoer);
 
         OscapScannerRemoteSsh();
         virtual ~OscapScannerRemoteSsh();
 
+        bool getUserIsSudoer() const;
         void setUserIsSudoer(bool userIsSudoer);
         virtual void setTarget(const QString& target);
         virtual void setSession(ScanningSession* session);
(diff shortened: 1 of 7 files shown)


src/OscapScannerRemoteSsh.cpp
```diff
@@ -328,6 +344,29 @@ void OscapScannerRemoteSsh::evaluate()
     signalCompletion(mCancelRequested);
 }
 
+void OscapScannerRemoteSsh::selectError(MessageType& kind, const QString& message)
```
matusmarhefka, line 347:
```
In file included from /home/jenkins/workspace/scap-workbench-pull-requests/src/MainWindow.cpp:24:
/home/jenkins/workspace/scap-workbench-pull-requests/include/OscapScannerRemoteSsh.h:48:33: error: ‘MessageType’ has not been declared
        virtual void selectError(MessageType& kind, const QString& message);
                                 ^~~~~~~~~~~
```

include/OscapScannerRemoteSsh.h
```diff
@@ -36,12 +36,18 @@ class OscapScannerRemoteSsh : public OscapScannerBase
...
+        void setUserIsSudoer(bool userIsSudoer);
         virtual void setTarget(const QString& target);
         virtual void setSession(ScanningSession* session);
 
         virtual QStringList getCommandLineArgs() const;
         virtual void evaluate();
 
+    protected:
+
+       virtual void selectError(MessageType& kind, const QString& message);
```
matusmarhefka, line 48:
```
In file included from /home/jenkins/workspace/scap-workbench-pull-requests/src/MainWindow.cpp:24:
/home/jenkins/workspace/scap-workbench-pull-requests/include/OscapScannerRemoteSsh.h:48:33: error: ‘MessageType’ has not been declared
        virtual void selectError(MessageType& kind, const QString& message);
                                 ^~~~~~~~~~~
```

matejak: This fails because of the dependency on #271

src/OscapScannerRemoteSsh.cpp
```diff
@@ -328,6 +344,29 @@ void OscapScannerRemoteSsh::evaluate()
...
+
+}
+
+void OscapScannerRemoteSsh::processError(QString& message)
+{
+    OscapScannerBase::processError(message);
+    if (mUserIsSudoer && message.contains(QRegExp("^sudo:")))
+    {
+        message.replace(QRegExp("^sudo:"), "Error invoking sudo on the host:");
+        message += ".\nOnly passwordless sudo setup on the remote host is supported by scap-workbench.";
```
matusmarhefka, line 366: As this is a special case what if we also add an additional short description how to do that?
```
Error invoking sudo on the host: a password is required. Only passwordless sudo setup on the remote host is supported by scap-workbench.
To configure a non-privileged user (for example 'oscap-user') to run only the oscap binary as root update the remote machine's 'sudoers' file or add a file like '/etc/sudoers.d/99-oscap-user' with:
# allow oscap-user to run openscap scanner
Defaults!/usr/bin/oscap !requiretty
oscap-user ALL=(root) NOPASSWD: /usr/bin/oscap
```
This will prolong the message but will point a user how to do it. Anyway it is just an idea, feel free to discard it and rather only extend the documentation.

matusmarhefka:
@matejak Overall this looks good. I have tested with both configs (passwordless sudo and sudo with passowrd) and it behaves as expected.

evgenyz:
Isn't it a security breach - allow to run `sudo oscap` without a password? It could also execute remediations (which could be anything).

matejak:
> Isn't it a security breach - allow to run `sudo oscap` without a password? It could also execute remediations (which could be anything).

I have thought of this as well. It is a much less of a risk than allowing root to perform those operations, but unfortunately we can't do more. So although it is not a regression, I see a missed opportunity to make scans safer.

Now what's the problem: `sudo` can restrict the command-line that it executes with superuser privileges. However, the sudoers file doesn't support regexes, so there is no way to tell to `sudo` that it should allow execution of any `oscap xccdf eval` command-line that doesn't include `--remediate`. Workbench supplies quite a lot of arguments to the remote program oscap, so one can't say that you expect `oscap xccdf eval <datastream>`. There is a report, oval results, and XCCDF/DS IDs passed along.

matusmarhefka:
I have tested the latest changes and everything works as expected. I agree that passwordless sudo is not ideal, but it is still better than running under root. I will merge this if there are no objections from @evgenyz

evgenyz:
No, I don't have any

PR #271, opened 2021-01-27
Refactored handling of scanner messages

matejak:
This PR refactors oscap message handling - it preserves the current functionality, while it improves handling of some errors and warnings that the scanner can produce.

- Better detection of errors and warnings using leading `E:` and `W:` that oscap emits.
- Possibility to expand the message filtering using inheritance.

### DIFF
--- a/include/OscapScannerBase.h
+++ b/include/OscapScannerBase.h
@@ -77,6 +77,21 @@ class OscapScannerBase : public Scanner
 
         ReadingState mReadingState;
 
+       enum MessageType
+       {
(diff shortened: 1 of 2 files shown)


matusmarhefka:
Nice change, I also tested on RHEL8 and the Diagnostics window now looks much more user friendly:

old:
Screenshot from 2021-01-22 16-16-38

and with the changes in this PR:
Screenshot from 2021-01-28 11-47-34

Thanks @matejak

Issue #262, opened 2020-06-01
Possible deference of a null pointer


matejak:
The `session` pointer that is deferenced may be null:

https://github.com/OpenSCAP/scap-workbench/blob/af528f8f7343d57ce93731e12ff95379cc4f2010/src/ResultViewer.cpp#L117

It's interesting that the `if (session)` block has not been used. Was that intentional?

PR #276, opened 2021-03-18
Fix potential NULL pointer dereference

cipherboy:
With the async behavior of Scanner (and its interactions in MainWindow),
it isn't obvious that Session can't be non-`NULL` here. Hence it is
probably better to move the session use into the if-guard.

Resolves: #262

`Signed-off-by: Alexander Scheel <alexander.m.scheel@gmail.com>`

---

Looks like @mpreisler originally wrote the if-guard but didn't mention why it was necessary either :-)


evgenyz:
LGTM

Issue #275, opened 2021-03-18
Higher resolution UI assets?


cipherboy:
Figured I'd open an issue in case anyone has any thoughts.

I have a 14" 1440p x1c 4th gen. When I load `scap-workbench` on Fedora 33, I notice a lot of blobby pixels in most of the icons. This is because they're all 16x16 and they're rendering in a space double the size (32x32).

I'm able to find direct replacements for the following assets from the tango project:

 - `benchmark.png` == `edit-paste`
 - `edit-undo.png` == `edit-undo`
 - `edit-redo.png` == `edit-redo`
 - `group.png` == `folder`
 - `value.png` == `text-x-generic`
 - `rule.png` == `preferences-system`

The following assets don't have direct replacements:

 - `collapsed-arrow.png`
 - `expanded-arrow.png`
 - `profile.png`

The two arrows are easy enough to recreate (Inkscape and create a quick triangle shape, matching color), but what about `profile.png`? Could it be replaced with something from tango like `x-office-spreadsheet`? 

Old:
![profile](https://user-images.githubusercontent.com/914030/111710462-20780680-8820-11eb-9e92-7d2efd8a713a.png)

Suggested:
![x-office-spreadsheet](https://user-images.githubusercontent.com/914030/111710771-abf19780-8820-11eb-9443-8a0cd9183c97.png)

I can also remove the pichart like so:
![x-office-spreadsheet-edited](https://user-images.githubusercontent.com/914030/111710940-11de1f00-8821-11eb-8711-d1dc33d975ba.png)

Lastly, what is everyone's preferences about size? I'd probably re-export at 64x64 or 128x128. I can directly use SVG assets as well (Tango provides such), but the issue is loading the customization takes a while if using SVGs. I'm not sure yet if 128x128 does the same; if it doesn't I'd prefer to go with that. Note that this only affects the source asset size, the file size in the GUI doesn't change.

Thoughts?

PR #277, opened 2021-03-24
Switch to high-quality assets

cipherboy:
Import SVGs from Tango project and re-export existing icons at 64x64 size. This should ensure they're sufficiently large for all display purposes.

Also included are three code changes:

 - Use SVG application icon instead of PNG; it is rarely loaded so there is no performance impact.
 - Mark `runwrapper.sh.in` executable, making `build/runwrapper.sh`
 - Set `AA_EnableHighDpiScaling`; see https://doc.qt.io/qt-5/highdpi.html.

Resolves: #275


cipherboy:
Interesting; this doesn't actually appear to resolve the issues on my laptop.

matejak:
The PR would probably benefit from a rebase, as it fails to compile on F34 for me.

PR #278, opened 2021-03-24
Use reference to avoid range-loop-construct diagnostic

cipherboy:
Pulled from scap-workbench-1.2.1-7 in Fedora Rawhide.

`Signed-off-by: Alexander Scheel <alex.scheel@canonical.com>`


cipherboy:
Unless anyone objects, I'm inclined to merge this directly since it was already committed to Rawhide and was authored by someone else. It seems like a trivial change.

evgenyz:
Sure, why not?

PR #279, opened 2021-03-25
Allow profile description customization

cipherboy:
When opening the profile editor, the description field is populated but
is marked `readOnly` for some reason; all the back-end code existed to
make this read-write (except in the case when the profile lacks a
description, see the comment in [`TailoringWindow::setProfileDescription`](https://github.com/OpenSCAP/scap-workbench/blob/v1-2/src/TailoringWindow.cpp#L529)
and [`TailoringWindow::setProfileTitle`](https://github.com/OpenSCAP/scap-workbench/blob/v1-2/src/TailoringWindow.cpp#L491)).

This makes the field read-write.

Note that the behavior is less than ideal: `<xccdf:description />` can
include HTML content, and while this is a HTML-aware editor, hitting the
return key inserts merely a '`\n`' into the text stream and not a `<br />`
(or better, `<xhtml:br />` with XML-NS awareness) element.

Perhaps this behavior should be corrected at the openscap library level,
detecting the HTML XML namespace applied on the `<xccdf:description />`
element and namespacing the input correctly?

`Signed-off-by: Alexander Scheel <alex.scheel@canonical.com>`

---

n.b.: I used qtcreator to modify the UI, let me know if you want me to go back and undo the other changes to the UI file and only modify the boolean.

### DIFF
--- a/ui/ProfilePropertiesDockWidget.ui
+++ b/ui/ProfilePropertiesDockWidget.ui
@@ -21,7 +21,16 @@
   </property>
   <widget class="QWidget" name="dockWidgetContents">
    <layout class="QVBoxLayout" name="verticalLayout">
-    <property name="margin">
+    <property name="leftMargin">
+     <number>0</number>
+    </property>
+    <property name="topMargin">
+     <number>0</number>
+    </property>
+    <property name="rightMargin">
+     <number>0</number>
+    </property>
+    <property name="bottomMargin">
      <number>0</number>
     </property>
     <item>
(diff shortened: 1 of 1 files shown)


yuumasato:
I think it makes sense to enable editing the `description` of a tailoring profile.
Couldn't find anything preventing this in the standard.

cipherboy:
Cool, thanks @yuumasato! Is there anything else? Or should I merge this tomorrow?

cipherboy:
Hearing nothing, and seeing as this is a minor change, I'm going ahead and merging. Thanks for the review @yuumasato! :)

PR #286, opened 2021-07-30
Do not set rpath

jan-cerny:
See: https://fedoraproject.org/wiki/Changes/Broken_RPATH_will_fail_rpmbuild
Similar to: https://github.com/OpenSCAP/openscap/pull/1765


matejak (approved):
LGTM, thanks!

PR #288, opened 2021-08-12
Fix appdata settings

henrich:
Update appdata related things as 

* .desktop file cannot use "-", so replace it as "_"
* not /usr/share/appdata but /usrshare/metainfo now
* appropriate appdata.xml


evgenyz (approved):

evgenyz:
Thank you!

PR #295, opened 2021-11-15
Fix appdata, desktop and icons for Linux (Freedesktop)

evgenyz:
Satisfy:
- https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html#install_icons
- https://freedesktop.org/software/appstream/docs/chap-Quickstart.html

`desktop-file-validate org.open_scap.scap_workbench.desktop` and `appstreamcli validate org.open_scap.scap_workbench.metainfo.xml` are happy.

Continuation of #288.


org.open_scap.scap_workbench.appdata.xml
```diff
@@ -1,10 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<component type="desktop">
-    <id>org.openscap.scap_workbench.desktop</id>
+<component type="desktop-application">
+    <id>org.open_scap.scap_workbench</id>
     <metadata_license>CC0-1.0</metadata_license>
     <project_license>GPL-3.0+</project_license>
     <name>SCAP Workbench</name>
     <summary>A GUI tool that serves as an SCAP scanner and provides tailoring functionality for SCAP content</summary>
+    <launchable type="desktop-id">org.open_scap.scap_workbench.desktop</launchable>
```
matejak, line 8: What about
```suggestion
    <launchable type="desktop-id">org.open_scap.scap_workbench.desktop</launchable>
    <content_rating/>
```

evgenyz: But what rating exactly? We are using downloadable content, so we can't know what could possibly be there.

matejak: It is true that workbench will display whatever the user supplies to it, and I can imagine `money-advertising` becoming an issue in the extreme realistic case, but with this approach you couldn't rate a text editor or media player.
Anyway, I don't want to get stuck here because of a validation info output.

matejak (approved):
LGTM, merging.

org.open_scap.scap_workbench.appdata.xml
evgenyz: That's pretty much the reason I left it unrated. It's better to not say anything than say something controversial.

evgenyz: And we, unlike a media player, supply a list of downloadable content, which ATM is under our control, but still technically not part of the package.

PR #298, opened 2022-01-28
Fix remote scanning with sudo

matejak:
This fix ensures that various temp files that are created and used in the process have the same owner all the time - if the scan is executed with sudo, then the owner of report, ARF and result files has to be root from the beginning to the end.

The root user in RHEL9 can't just write to files owned by other users.

This PR fixes https://bugzilla.redhat.com/show_bug.cgi?id=2047740


matejak:
So this PR introduces more problems, e.g. need to configure sudo with `mktemp`, `cat` and `rm` programs, which doesn't sound good at all. It turns out that the problem is specific to `oscap` rather than to root restrictions on RHEL9.

PR #306, opened 2022-12-19
Update workflow to point to main branch and replace words with `leader`

ggbecker:


src/RemoteSsh.cpp
```diff
@@ -131,13 +131,13 @@ void SshConnection::connect()
         args.append(SCAP_WORKBENCH_LOCAL_SSH_PATH);
 #endif
 
-        args.append("-M"); // place ssh client into "master" mode for connection sharing
+        args.append("-M"); // place ssh client into "leader" mode for connection sharing
```
yuumasato, line 134: I found no evidence of a Leader mode in SSH.

```suggestion
        args.append("-M"); // place ssh client in control of the shared connection socket
```

src/RemoteSsh.cpp
```diff
@@ -110,7 +110,7 @@ void SshConnection::connect()
         }
 
         mSocketDir = new SpacelessQTemporaryDir();
-        mMasterSocket = mSocketDir->path() + "/ssh_socket";
+        mLeaderSocket = mSocketDir->path() + "/ssh_socket";
```
yuumasato, line 113: I think `Main` fits better than `Leader`.
Or how about using `control`? `mControlSocket`

yuumasato (changes requested):

src/RemoteSsh.cpp
ggbecker: In this case control seems better.. let me update

src/RemoteSsh.cpp
ggbecker: will update it accordingly

yuumasato (approved):

